PR #5, opened 2022-07-27
First table migration for testing purposes

arivepr:
Adding this just for testing purposes.

Proving the service will actually create the tables when running for the first time, otherwise pulling them in. 

Screen Shot 2022-07-27 at 10 44 24 AM

### DIFF
--- a/rest/database/db.go
+++ b/rest/database/db.go
@@ -3,6 +3,7 @@ package database
 import (
 	"fmt"
 
+	"github.com/RedHatInsights/chrome-service-backend/rest/models"
 	"github.com/RedHatInsights/chrome-service-backend/config"
 	"gorm.io/gorm"
 
@@ -28,6 +29,10 @@ func Init() {
 	dialector = postgres.Open(dbdns)
 
 	DB, err = gorm.Open(dialector, &gorm.Config{})
+
+	if !DB.Migrator().HasTable(&models.FavoritePage{}) {
+		DB.Migrator().CreateTable(&models.FavoritePage{})
+	}
 
 	if err != nil {
 		panic(fmt.Sprintf("Database connection failed: %s", err.Error()))
--- a/rest/database/db.go
+++ b/rest/database/db.go
@@ -3,10 +3,10 @@ package database
 import (
 	"fmt"
 
-	"github.com/RedHatInsights/chrome-service-backend/rest/models"
 	"github.com/RedHatInsights/chrome-service-backend/config"
-	"gorm.io/gorm"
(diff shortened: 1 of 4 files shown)


rest/models/base.go
```diff
@@ -2,11 +2,13 @@ package models
...
   "time"
+  "gorm.io/gorm"
 )
 
 // Generic Struct used throughout models in this service.
 type BaseModel struct {
   Id uint
   CreatedAt time.Time
   UpdatedAt time.Time
+  DeletedAt gorm.DeletedAt
```
Hyperkid123, line 13: Is it possible to use `time.Time` as well? The issue with the gorm type is that indexes the field as well and causes issue with OpenAPI within spec.

arivepr: For sure! if I understand this correctly, you want me to go from gorm.DeletedAt, to just a simple time stamp as well?

rest/database/db.go
```diff
@@ -29,6 +28,17 @@ func Init() {
...
 	DB, err = gorm.Open(dialector, &gorm.Config{})
 
+	// Migration/Creation of data tables for DB
+	if !DB.Migrator().HasTable(&models.UserIdentity{}) {
+		DB.Migrator().CreateTable(&models.UserIdentity{})
+	}
+	if !DB.Migrator().HasTable(&models.FavoritePage{}) {
+		DB.Migrator().CreateTable(&models.FavoritePage{})
+	}
+	if !DB.Migrator().HasTable(&models.LastVisitedPage{}) {
```
Hyperkid123, line 38: This model does not exist yet.

rest/database/db.go
```diff
@@ -29,6 +28,17 @@ func Init() {
 
 	DB, err = gorm.Open(dialector, &gorm.Config{})
 
+	// Migration/Creation of data tables for DB
+	if !DB.Migrator().HasTable(&models.UserIdentity{}) {
+		DB.Migrator().CreateTable(&models.UserIdentity{})
```
Hyperkid123, line 33: This model does not exist yet

rest/database/db.go
arivepr: Yeah this file change went past me when I stashed the other changes, I'm changing it back rn

arivepr:
Please excuse the Merge, had to wrestle git for a bit 😅 😅 @Hyperkid123

rest/models/base.go
Hyperkid123: Yeah. The gorm.DeletedAt is a bit problematic. We will see how it turns out.

PR #6, opened 2022-08-01
Adding additional models

arivepr:
Addresses [this jira card](https://issues.redhat.com/browse/RHCLOUD-20613)

Marked as draft given that I noticed the models in quickstarts had a ton more gorm `details`, wanted to make sure I wasn't missing more stuff on that end (I don't know much of gorm, nor what we're doing with those things on the quickstarts side)

Here's a snapshot of the created tables: 
602B8A49-85C4-45EA-B294-7E9EFD9D62A8_4_5005_c

SQL points at the correct properties being created at first-run time

### DIFF
--- a/rest/models/LastVisitedPage.go
+++ b/rest/models/LastVisitedPage.go
@@ -1,0 +1,7 @@
+package models
+
+type LastVisitedPage struct {
+  BaseModel
+  Pathname string
+  Title string
+}
--- a/rest/models/LastVisitedPage.go
+++ b/rest/models/LastVisitedPage.go
@@ -4,4 +4,5 @@ type LastVisitedPage struct {
   BaseModel
   Pathname string `json:"pathname"`
   Title string    `json:"title"`
+  UserId uint     `json:"userId"`
 }
--- a/rest/models/LastVisitedPage.go
+++ b/rest/models/LastVisitedPage.go
@@ -2,6 +2,6 @@ package models
 
 type LastVisitedPage struct {
   BaseModel
-  Pathname string
-  Title string
+  Pathname string `json:"pathname"`
+  Title string    `json:"title"`
 }
--- a/rest/models/UserIdentity.go
+++ b/rest/models/UserIdentity.go
@@ -1,0 +1,14 @@
+package models 
+import (
+  "time"
+)
+
+type UserIdentity struct {
+  BaseModel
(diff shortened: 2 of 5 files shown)


rest/models/UserIdentity.go
```diff
@@ -0,0 +1,14 @@
...
+  "time"
+)
+
+type UserIdentity struct {
+  BaseModel
+  AccountId string
+  FirstLoging bool
+  DayOne bool
+  LastLogin time.Time
+  LastVisitedPages []LastVisitedPage `gorm:"many2many:user_last_visited_pages"`
```
Hyperkid123, line 12: These will not be many to many relations. Each last visited page entry will be only for one specific user. Same goes for favorite pages.

It should be one to many relation.

rest/models/LastVisitedPage.go
```diff
@@ -0,0 +1,7 @@
+package models
+
+type LastVisitedPage struct {
+  BaseModel
+  Pathname string
```
Hyperkid123, line 5: We also have to add the `json` definition. So go knows how to marshal/unmarshal JSON

arivepr: Should I add that json definition to all of the properties? 
so, i.e: Pathname string `json: pathname` and so on for the rest?

Hyperkid123: Yeah

rest/models/UserIdentity.go
```diff
@@ -0,0 +1,14 @@
...
+  "time"
+)
+
+type UserIdentity struct {
+  BaseModel
+  AccountId         string              `json:"accountId,omitempty"`
+  FirstLogin        bool                `json:"firstLogin"`
+  DayOne            bool                `json:"dayOne"`
+  LastLogin         time.Time           `json:"lastLogin"`
+  LastVisitedPages  []LastVisitedPage   `gorm:"one2many:user_last_visited_pages"`
```
Hyperkid123, line 12: This is how the one2many should be defined: https://gorm.io/docs/has_many.html

arivepr: Yes! was hacking on my end, fix is up now. Let me know if that works for you 🧑‍🍳

Hyperkid123 (approved):

PR #11, opened 2022-08-15
Adding GET endpoint for UserEntity

arivepr:
Adresses [this JIRA card.](https://issues.redhat.com/browse/RHCLOUD-20628)

### DIFF
--- a/rest/routes/identity.go
+++ b/rest/routes/identity.go
@@ -10,17 +10,14 @@ import (
   "github.com/go-chi/chi/v5"
 )
 
+// Use the user obj in context to pull full data row from DB
 func GetUserIdentity(w http.ResponseWriter, r *http.Request) {
-  user := r.Context().Value(util.USER_CTX_KEY).(models.UserIdentity) 
-  userID := user.ID
-  response := make(map[string]models.UserIdentity)
-  
-  userData, err := service.GetUserIdentityByID(userID)
+  user := r.Context().Value(util.USER_CTX_KEY).(models.UserIdentity)
+  updatedUser, err := service.GetUserIdentityData(user)
   if err != nil {
     panic(err)
   }
-  response["data"] = userData
-  json.NewEncoder(w).Encode(response)
+  json.NewEncoder(w).Encode(updatedUser)
 }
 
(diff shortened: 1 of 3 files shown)


arivepr:
So dismissed updating the `last_login` with the Get endpoint for this PR since I wasn't sure if by simply hitting the Get endpoint we know that the user login.

I can append another endpoint that specifically addresses this by doing something like `getUserForLogin` or whatever, and just add the correct timestamp there. Not sure though what the best approach is there though 😅

rest/routes/identity.go
```diff
@@ -0,0 +1,28 @@
...
+func GetUserIdentity(w http.ResponseWriter, r *http.Request) {
+  user := r.Context().Value(util.USER_CTX_KEY).(models.UserIdentity) 
+  userID := user.ID
+  response := make(map[string]models.UserIdentity)
+  
+  userData, err := service.GetUserIdentityByID(userID)
+  if err != nil {
+    panic(err)
+  }
+  response["data"] = userData
```
Hyperkid123, line 22: The response should just return the `userData` directly. If a response is not a list, we just return the entity.

rest/routes/identity.go
```diff
@@ -0,0 +1,28 @@
...
+  "github.com/RedHatInsights/chrome-service-backend/rest/models"
+  "github.com/go-chi/chi/v5"
+)
+
+func GetUserIdentity(w http.ResponseWriter, r *http.Request) {
+  user := r.Context().Value(util.USER_CTX_KEY).(models.UserIdentity) 
+  userID := user.ID
+  response := make(map[string]models.UserIdentity)
+  
+  userData, err := service.GetUserIdentityByID(userID)
```
Hyperkid123, line 18: The context under `util.USER_CTX_KEY` actually gives you the user identity already. It does a `FirstOrCreate` lookup, that will either retrieve the entity or creates it. The logic is in the `InjectUser` middleware.

Essentially you are doing the same query twice.

rest/routes/identity.go
arivepr: Just changed it on my end, reached out to ya about the other item 😎

Hyperkid123 (approved):

PR #17, opened 2022-11-29
Fix file permissions for container process

BlakeHolifield:
In order to write the certificate we get from Clowder, we need to fix the permissions on the Dockerfile.

* Update permission for `/app` where the service will run
* Remove debug for database connection so we don't leak passwords
* Remove binaries from the repo and add them to `.gitignore`


ryelo (approved):
Beautiful PR

PR #28, opened 2023-02-28
Do not send nil as visited bundles.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-24446

### Changes
- Do not return visitedBundles attribute if value is equal to `null`
- Add migration script to fill `visited_bundles` to `{}` if record value is equal to `NULL`
- Run DB migrations in a single transaction

### DIFF
--- a/cmd/migrate/migrate.go
+++ b/cmd/migrate/migrate.go
@@ -6,15 +6,38 @@ import (
 	"github.com/RedHatInsights/chrome-service-backend/rest/models"
 	"github.com/joho/godotenv"
 	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
 )
 
 func main() {
 	godotenv.Load()
 	config.Init()
 	database.Init()
-	err := database.DB.AutoMigrate(&models.FavoritePage{}, &models.LastVisitedPage{}, &models.SelfReport{}, &models.UserIdentity{}, &models.ProductOfInterest{})
+
+	var bundleRes *gorm.DB
+
+	// Wrap DB migration into a single transaction
+	err := database.DB.Transaction(func(tx *gorm.DB) error {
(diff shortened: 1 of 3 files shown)


cmd/migrate/migrate.go
```diff
@@ -6,15 +6,38 @@ import (
...
+	var bundleRes *gorm.DB
+
+	// Wrap DB migration into a single transaction
+	err := database.DB.Transaction(func(tx *gorm.DB) error {
+		err := tx.AutoMigrate(&models.FavoritePage{}, &models.LastVisitedPage{}, &models.SelfReport{}, &models.UserIdentity{}, &models.ProductOfInterest{})
+		if err != nil {
+			return err
+		}
+		// Migrate all nil visited_bundles to empty object
+		bundleRes = database.DB.Model(&models.UserIdentity{}).Where("visited_bundles IS NULL").Update("visited_bundles", []byte(`{}`))
```
karelhala, line 26: Wouldn't it be better here to create new empty record? We already know the model `&models.UserIdentity{}` so we can use it as `json.Marshal(newModel)` instead of `[]byte('{}')`. What ifthe `UserIdentity` changes and will have some data required, we'd have to update the `Update` function as well.

karelhala:
```suggestion
                model = &models.UserIdentity{}
                b, err := json.Marshal(model)
                if err != nil {
                    return err
                }
		bundleRes = database.DB.Model(model).Where("visited_bundles IS NULL").Update("visited_bundles", b)
```

karelhala (approved):
Looking good, just a minor nit pick.

cmd/migrate/migrate.go
Hyperkid123: I am not sure if this is possible. Creating an empty model does not create an empty JSON for the visitedBundles attribute. It does assign `nil` to the visited_bundles. And in your example, you are also setting the `visited_bundles` with a new empty model.

I don't think this is doable without iterating over all existing records.

PR #37, opened 2023-04-06
[RHCLOUD-24573] Entrypoints for learning resources

BlakeHolifield:


static/.prettierrc.json
Hyperkid123, line 1: I don't think prettier will work without having nodejs setup. (And we don't want that). Unless you have it installed globally o as an IDE extension.

Hyperkid123 (approved):

PR #38, opened 2023-04-11
Add all services links to all environments

karelhala:
### Description

Since chrome is not using the services config anywhere (it will use it in stage once https://github.com/RedHatInsights/insights-chrome/pull/2398 is merged) we can safely copy what's in stage to every environment. If the service is not available in the navigation config it won't be added to the final list anyway.

### DIFF
--- a/static/beta/prod/services/services.json
+++ b/static/beta/prod/services/services.json
@@ -1,1 +1,217 @@
-[]
+[
+  {
+    "id": "applicationServices",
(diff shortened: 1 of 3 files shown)


Hyperkid123:
@karelhala I think you are missing stage beta env. it should be the same as prod beta right now.

Hyperkid123 (approved):

PR #44, opened 2023-04-12
Add nav link to prod-beta env for Cost Management

dlabrecq:
Add optimizations nav link to prod-beta env for Cost Management

https://issues.redhat.com/browse/COST-3600


Hyperkid123 (commented):
@karelhala does this match with the IDs for the new all services data?

karelhala (approved):

karelhala:
@Hyperkid123 yes, I haven't migrated other envs to use the IDs, but this is the same as what we have in stage config.

PR #47, opened 2023-04-12
Github Action workflow to validate fed-modules.json

riprasad:
In the recent attempt to add a new application, we encountered the issue of not using camel casing once again for the app-id. This has happened multiple times before and we only realize the mistake after spending a considerable amount of time on debugging.

[https://github.com/RedHatInsights/chrome-service-backend/pull/45](https://github.com/RedHatInsights/chrome-service-backend/pull/45 )

To address this, I am proposing the addition of a new Github Action workflow that will be triggered only when there is a change in fed-modules.json in any subdirectory of the static folder. The workflow will perform a validation check to ensure that the app-id specified in the fed-modules.json files is using camel case.This will ensure uniformity and save people from spending unnecessary time debugging the issue.

### DIFF
--- a/.github/scripts/validate-fed-modules.sh
+++ b/.github/scripts/validate-fed-modules.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+
+# https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
+set -euo pipefail
+
+
+# Find all files named "fed-modules.json" in the "static" directory
+# and store their paths in an array named "files".
+files=($(find static -name "fed-modules.json"))
+
+valid=true
+
+
+for file in "${files[@]}"
+do
+
+  # Read the contents of the file and pass them to the jq command to extract all keys that are not camel-cased.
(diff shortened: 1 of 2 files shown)


riccardo-forina (commented):
Have my 💯 !

Hyperkid123:
@riprasad  The ids do not have to be in camel case. They have to match whatever id is used in your build. I can contain any characters besides `-`. And even this restriction will go away in a near future.

riccardo-forina:
The problem is that even if our id uses dash-case in the insights appname property, it gets converted to lowerCamelCase by the fed module plugin without any option to change that behavior. This has bitten us at least three times already in the past.

Hyperkid123:
@riccardo-forina that is because of the restriction of the `-` character which is a webpack limitation. It is mentioned in the guides [here](https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/docs/chrome/docs.md#registering-new-module-app-to-chrome).

This limitation does not apply to modules exposing themselves as `jsonp` library type which will become the default in a near future, Some modules (hac) are already using it. We cannot restrict IDs to be camel case only.

riccardo-forina:
Gotcha. Thanks for clarifying that.

PR #52, opened 2023-04-17
Enable quay UI in stage beta.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-25311

Have to verify once the stage env is back up

### DIFF
--- a/static/beta/stage/modules/fed-modules.json
+++ b/static/beta/stage/modules/fed-modules.json
@@ -963,5 +963,22 @@
                 ]
             }
         ]
+    },
+    "quay-ui-plugin": {
+        "manifestLocation": "/apps/quay/plugin-manifest.json",
+        "modules": [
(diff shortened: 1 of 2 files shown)


static/beta/stage/modules/fed-modules.json
```diff
@@ -963,5 +963,22 @@
...
+            {
+                "id": "quay",
+                "module": "QuayPluginMain",
+                "routes": [
+                    {
+                        "pathname": "/containers"
+                    },
+                    {
+                        "pathname": "/quay"
+                    }
```
karelhala, line 979: Do we need this nav as well?

Hyperkid123: It should be `/containers/quay`

karelhala (approved):

PR #57, opened 2023-04-19
First attempt at second attempt to get Trusted Content configured.

bobmcwhirter:
Honestly, I have no idea what I'm doing here.

### DIFF
--- a/static/beta/stage/modules/fed-modules.json
+++ b/static/beta/stage/modules/fed-modules.json
@@ -980,5 +980,22 @@
                 ]
             }
         ]
+    },
+    "trust": {
+        "manifestLocation": "/apps/trust/fed-mods.json",
+        "modules": [
+            {
+                "id": "trust",
+                "module": "./RootApp",
+                "routes": [
+                    {
+                        "pathname": "/inventory"
+                    },
+                    {
+                        "pathname": "/package"
+                    }
+                ]
+            }
+        ]
     }
 }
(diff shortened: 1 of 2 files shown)


static/beta/stage/modules/fed-modules.json
```diff
@@ -980,5 +980,22 @@
...
+    },
+    "trust": {
+        "manifestLocation": "/apps/trust/fed-mods.json",
+        "modules": [
+            {
+                "id": "trust",
+                "module": "./RootApp",
+                "routes": [
+                    {
+                        "pathname": "/inventory"
```
Hyperkid123, line 992: This should be a full pathname to the location where the module should be mounted. I don't think `/insights` is the place. What is the URL to the page where these modules should be available at?

static/beta/stage/navigation/trust-navigation.json
```diff
@@ -0,0 +1,11 @@
+{
+    "id": "trust",
+    "title": "Trusted Content",
+    "navItems": [
+        {
+            "title": "Trusted Content",
+            "appId": "trust",
+            "href": "/trust"
```
Hyperkid123, line 8: Again this URL does not seems correct. The link href should include the module route.

Hyperkid123 (commented):
Its almost correct :slightly_smiling_face: we just need to fire out the routing and links.

karelhala:
Closing because of inactivity, if this is still required please open a new PR.

PR #68, opened 2023-04-25
[prod-beta] Add missing fed-modules entry for HCS

dlabrecq:
1. Add missing fed-modules entry
2. Omit permissions check for the "Overview" link.

"The intent was to hide the HCS UI completely from non-HCS users" -- Greg Bowman

### DIFF
--- a/static/beta/prod/modules/fed-modules.json
+++ b/static/beta/prod/modules/fed-modules.json
@@ -751,6 +751,20 @@
             }
         ]
     },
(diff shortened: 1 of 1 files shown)


karelhala:
@dlabrecq thank you for the PR, I completely forgot to add the HCS to fed-modules. Why do we want to remove the permissions check? If we were to remove it, the item will be visible both in all-services and in navigation.

karelhala (approved):

PR #89, opened 2023-05-02
Add Quay module to stage stable

syed:
validated on beta stage and is ready to be released to stable

### DIFF
--- a/static/stable/stage/modules/fed-modules.json
+++ b/static/stable/stage/modules/fed-modules.json
@@ -894,7 +894,7 @@
               ]
           }
       ]
-  }
+  },
   "quay-ui-plugin": {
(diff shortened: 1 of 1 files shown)


Hyperkid123 (approved):

static/stable/stage/modules/fed-modules.json
```diff
@@ -895,4 +895,21 @@
           }
       ]
   }
+  "quay-ui-plugin": {
```
Hyperkid123, line 898: I think you are missing a `,` before the new entry.

syed: thanks. Fixed it

Hyperkid123 (approved):

PR #104, opened 2023-05-04
RHIF-190: rename Overview to Dashboard, Inventory to Systems navs in stage env

mkholjuraev:
This renames the following navigations in STAGE env to be aligned with the new Navigation changes [RHIF-190](https://issues.redhat.com/browse/RHIF-190).

Overview => Dashboard (navigation id is also changed to dashboard to make it same is the app URL)
Inventory => Systems

I will create the same PR for prod env once gets verified and ack.

### DIFF
--- a/static/stable/stage/navigation/rhel-navigation.json
+++ b/static/stable/stage/navigation/rhel-navigation.json
@@ -3,9 +3,9 @@
     "title": "Red Hat Insights",
     "navItems": [
       {
-        "id": "overview",
+        "id": "dashboard",
         "appId": "dashboard",
-        "title": "Overview",
+        "title": "Dashboard",
         "filterable": false,
         "href": "/insights/dashboard",
         "product": "Red Hat Insights"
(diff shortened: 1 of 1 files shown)


mkholjuraev:
@Hyperkid123 A quick is appreciated for this important PR.

static/stable/stage/navigation/rhel-navigation.json
```diff
@@ -3,9 +3,9 @@
     "title": "Red Hat Insights",
     "navItems": [
       {
-        "id": "overview",
+        "id": "dashboard",
```
Hyperkid123, line 6: This changes has to be reflected in `service.json` if the link is supposed to show up in the new dropdown

mkholjuraev: Thank you! Do you mean all changes or only the change of nav id?

mkholjuraev: Went through service.json under /stage/services folder. 

1. As the dashboard is not a real service but an overall page for RHEL insights, it seems we do not have to make any changes for Overview => dashboard change 
2. The inventory is renamed already inside the dropdown without changing it's ID. This also seems not needed to be reflected in /stage/services/servises.json file.

Let me know if I miss anything, it is my first PR in this new repo :)

Hyperkid123 (approved):

PR #106, opened 2023-05-04
RHIF-190: undo insights dashboard nav ID change

mkholjuraev:
It seems changing the ID of the RHEL dashboard nav was not a good idea. After the PR got merged, this renaming did get reflected. This undoes that ID change.

Small question if you have time to answer, otherwise please ignore it: 
I am wondering what is difference between rhel-navigation.json and insights-navigation.json? It seems they are mapped to each other?

### DIFF
--- a/static/beta/stage/navigation/rhel-navigation.json
+++ b/static/beta/stage/navigation/rhel-navigation.json
@@ -5,7 +5,7 @@
     {
       "id": "overview",
       "appId": "dashboard",
-      "title": "Overview",
+      "title": "Dashboard",
       "filterable": false,
       "href": "/insights/dashboard",
       "product": "Red Hat Insights"


mkholjuraev:
@Hyperkid123 can ask your one more review on this?

Hyperkid123:
> Small question if you have time to answer, otherwise please ignore it:
I am wondering what is difference between rhel-navigation.json and insights-navigation.json? It seems they are mapped to each other?

@mkholjuraev there is/was a discussion about renaming the insights bundle to Rhel. Because this question was popping out quote often, instead of duplicating/renaming the nav file, we just created a symlink so the devs don't have to care.

Hyperkid123 (approved):

PR #129, opened 2023-05-09
[RHCLOUD-25832] Enable websockets 

BlakeHolifield:
* Setup local kafka env and helper tools
* Enable basic broadcast through the API
* Setup logrus to avoid spamming kafka reader messages
* Logrus now supports one of `info | debug` as a `LOG_LEVEL` 
* Adds feature flag support with fallback when the unleash service cannot be contacted.
* Refactors dependencies in to `init()`
* Adds docs for feature flags


Hyperkid123:
@BlakeHolifield can we use the feature flags to make sure the topic is used only in stage env? There will be some deployments to prod even during code freeze and I don't want to risk anything.

BlakeHolifield:
> @BlakeHolifield can we use the feature flags to make sure the topic is used only in stage env? There will be some deployments to prod even during code freeze and I don't want to risk anything.

Yep! I'll get that setup. Good call.

deploy/clowdapp.yml
```diff
@@ -100,9 +103,24 @@ objects:
...
+      INTERCOM_HAC_CORE: dGVzdFZhbHVl
+      INTERCOM_OPENSHIFT_DEV: dGVzdFZhbHVl
+      SEARCH_CLIENT_SECRET_PROD: dGVzdFZhbHVl
+      SEARCH_CLIENT_SECRET_STAGE: dGVzdFZhbHVl
+    type: Opaque
+
 parameters:
 - name: LOG_LEVEL
-  value: INFO
+  value: Inggo
```
Hyperkid123, line 123: Is this change intentional?

BlakeHolifield: Not even a little bit. I'll fix it.

Hyperkid123 (approved):

BlakeHolifield:
We're going to keep this on hold until after summit.

Hyperkid123:
Converting to draft to prevent accidental clicking the :eject_button:

PR #131, opened 2023-05-10
RHCLOUD-25835 Offboard RHOSAK

josejulio:
RHOSAK is not sending notifications anymore. It was the only application part of the `Application Services` bundles thus we are removing it from the menus.

### DIFF
--- a/static/stable/prod/navigation/user-preferences-navigation.json
+++ b/static/stable/prod/navigation/user-preferences-navigation.json
@@ -7,36 +7,6 @@
             "appId": "userPreferences",
             "title": "Email preferences",
             "href": "/user-preferences/email"
-        },
-        {
-            "title": "Notifications",
-            "expandable": true,
-            "routes": [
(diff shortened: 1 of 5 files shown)


static/stable/prod/navigation/user-preferences-navigation.json
```diff
@@ -7,36 +7,6 @@
...
-                    "title": "Red Hat Enterprise Linux",
-                    "href": "/user-preferences/notifications/rhel"
-                },
-                {
-                    "id": "console",
-                    "appId": "userPreferences",
-                    "title": "Console",
-                    "href": "/user-preferences/notifications/console"
-                }
-            ]
```
josejulio, line 39: This was off-sync from preview/beta. The submenus are no longer used right now. I could remove this off this PR if its intended to be fixed by other PR.

Hyperkid123 (approved):

PR #138, opened 2023-05-18
add service-accounts app

riccardo-forina:


riccardo-forina:
This is just a preview for the app here https://github.com/riccardo-forina/service-accounts

PR #145, opened 2023-05-19
[RHCLOUD-25868] Nav linking from /openshift/insights/advisor/clusters to /openshift

rsun19:
- The "clusters" link under infrastructure currently links to `/openshift/insights/advisor/clusters`, but now it's been changed to link to `/openshift`.
- However, when I try to access it locally, it throws this error shown below. However, it returns the correct end point.
image

- This also occurs when I try to access `/openshift` through containers, and the endpoints `/openshift/overview` and `/openshift/downloads`.
- Would appreciate a second look, since inputting the link directly works on console.redhat.com, but not when navigating it through chrome-service-backend.
- So the error thrown might be an OpenShift issue specifically.

### DIFF
--- a/static/beta/prod/navigation/openshift-navigation.json
+++ b/static/beta/prod/navigation/openshift-navigation.json
@@ -62,7 +62,7 @@
                             "description": "See targeted recommendations to optimize your OpenShift clusters’ availability, performance, and security."
                         },
                         {
-                            "id": "clusters",
+                            "id": "advisor-clusters",
                             "appId": "ocpAdvisor",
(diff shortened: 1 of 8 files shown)


Hyperkid123:
@rsun19 these are network errors that are triggering the development override (can be closed with the close button). Can you tell me in which environment have you been trying this locally? It might be just a local env issue with the endpoint. Especially if you were trying it locally in stage env. (OpenShift UI does not work in stage environment dues to their auth requirements.)

Hyperkid123 (approved):

PR #147, opened 2023-05-23
Use cloud events envelope

Hyperkid123:
Part of: https://issues.redhat.com/browse/RHCLOUD-26019

### DIFF
--- a/rest/cloudEvents/cloudEvents.go
+++ b/rest/cloudEvents/cloudEvents.go
@@ -1,0 +1,62 @@
+package cloudEvents
+
+import (
+	"fmt"
+	"time"
+)
+
+// Cloud events spec: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md
+
+type DataContentType string
+type SpecVersion string
+
+const (
+	ApplicationJson DataContentType = "application/json"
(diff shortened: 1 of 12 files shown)


rest/featureflags/featureflags.go
```diff
@@ -18,6 +19,7 @@ type FFClient struct {
 var ffClient *FFClient
 
 func newClientWrapper(cfg *config.ChromeServiceConfig) (*unleash.Client, error) {
+	fmt.Println(cfg)
```
BlakeHolifield, line 22: Do you still need this?

rest/cloudEvents/cloudEvents.go
```diff
@@ -0,0 +1,68 @@
+package cloudEvents
```
BlakeHolifield, line 1: I think the standard is no mixed case for go, so `cloudevents` would be fine

BlakeHolifield (changes requested):
Short feedback and small details

rest/featureflags/featureflags.go
Hyperkid123: Nope

BlakeHolifield (approved):

PR #148, opened 2023-05-23
[RHCLOUD-26017] [Chrome service] Add navigation file unique ID validation

rsun19:
Added validation to make sure that no id's are repeated. It has already detected a duplicate in the itless folder. Let me know if the code needs to be cleaner, or if you don't want the program to throw an exception if a duplicate is found!

It uses recursion to access the nested ids in the json file.

### DIFF
--- a/cmd/validate/validateNavigation.go
+++ b/cmd/validate/validateNavigation.go
@@ -33,7 +33,7 @@ func validateNavigation(cwd string) {
 
 		var data map[string]interface {}
 		
-		var arrayData [1] map[string]interface {};
+		var arrayData [] map[string]interface {};
 
 		fileContent, status := ioutil.ReadFile(file)
 		handleErr(status)
@@ -41,12 +41,18 @@ func validateNavigation(cwd string) {
 		ok:= json.Unmarshal(fileContent, &data)
 		if ok == nil{
 			//Note, only works if there are no errors unmarshalling!
-			parsingSetup(data, file)
+			duplicateCounter = make(map[string]int)
+			parseJSONIDs(data, file)
 		} else {
 			ok := json.Unmarshal(fileContent, &arrayData)
 			if ok == nil {
-				rootJSONFile := arrayData[0]
-				parsingSetup(rootJSONFile, file)
+				duplicateCounter = make(map[string]int)
+				jsonArrayData := make([]interface{}, len(arrayData))
+				for k,v := range arrayData {
+					jsonArrayData[k] = v
+				}
+				loopOverFields(jsonArrayData, file)
+				//parsingSetup(rootJSONFile, file)
 			} else {
 				panic(ok.Error())
 			}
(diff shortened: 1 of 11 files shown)


rsun19:
I think that checks fail because the program panics() when duplicates are found

cmd/validate/validateNavigation.go
```diff
@@ -15,6 +18,18 @@ func validateNavigation(cwd string) {
 	for _, file := range modulesFiles {
 		documentLoader := gojsonschema.NewReferenceLoader(fmt.Sprintf("file://%s", file))
 		result, err := gojsonschema.Validate(schemaLoader, documentLoader)
+		
+		var data map[string]interface {}
+		
+		fileContent, status := ioutil.ReadFile(file)
+		handleErr(status)
+
+		ok:= json.Unmarshal(fileContent, &data)
```
Hyperkid123, line 27: We should panic if there is a marshaling error. It does mean there is something wrong with the format.

cmd/validate/validateNavigation.go
```diff
@@ -15,6 +18,18 @@ func validateNavigation(cwd string) {
 	for _, file := range modulesFiles {
 		documentLoader := gojsonschema.NewReferenceLoader(fmt.Sprintf("file://%s", file))
 		result, err := gojsonschema.Validate(schemaLoader, documentLoader)
+
```
Hyperkid123, line 21: We should not skip the error check. Each block of code should have an escape hatch if it errors.

cmd/validate/validateNavigation.go
```diff
@@ -28,3 +43,81 @@ func validateNavigation(cwd string) {
...
+						// fmt.Println(id)
+					}
+				}
+				checkForNestedItems(file, navItem)
+			} 
+		}
+	}
+}
+
+func checkForNestedItems(file string, navItem map[string]interface{}) {
```
Hyperkid123, line 77: The `checkForNestedItems` seems to be almost identical to the `parseJSONIDs`. The only difference are the nested items key. Can we merge these two functions into one?

Also, the two branches within the if statement are very similar, I think these can be extracted into a function.

Hyperkid123 (changes requested):
You are correct about the CI failure

```
23:22:48 #13 1.236 panic: The id rhel in /src/chrome-service-backend/static/beta/prod/navigation/insights-navigation.json is not valid because it is duplicated
23:22:48 #13 1.236 
23:22:48 #13 1.236 
```

We should change the ID(s). I'll let you know the value(s).

rsun19:
I made the code cleaner and added more error throwing.

There were some things I didn't catch from before. For example, there was a root nested inside an array: i.e. [{id: ...}]

That is handled now. This algorithm also assumes that group id does not need to be recorded.

rsun19:
Made final commits, can now handle cases where the root is an array rather than a map. Code is way more concise.

Hyperkid123 (commented):
@rsun19 OK. This looks better. We now have to deal with the duplicate `rhel` id.

We have to do the following:
- Rename `rhel` to `subs.rhel` onside all `insights-navigation.json` files. Do not change the top-level id. Only the id attribute in the subscriptions section.

Then we should be good.

Hyperkid123 (approved):

PR #154, opened 2023-05-31
[RHCLOUD-26033] Validate incoming WS payload to match with cloud-events

rsun19:
Let me know if there needs to be more fields that needs to be validated.

### DIFF
--- a/main.go
+++ b/main.go
@@ -57,7 +57,7 @@ func main() {
 
 	// We might want to setup some event listeners at some point, but the pod will
 	// have to restart for these to take effect. We can't enable and disable websockets on the fly
-	if true || featureflags.IsEnabled("chrome-service.websockets.enabled") {
+	if featureflags.IsEnabled("chrome-service.websockets.enabled") {
 		// start the connection hub
 		go connectionhub.ConnectionHub.Run()
 		logrus.Infoln("Enabling WebSockets")
--- a/main.go
+++ b/main.go
@@ -57,7 +57,7 @@ func main() {
 
 	// We might want to setup some event listeners at some point, but the pod will
 	// have to restart for these to take effect. We can't enable and disable websockets on the fly
-	if featureflags.IsEnabled("chrome-service.websockets.enabled") {
+	if true || featureflags.IsEnabled("chrome-service.websockets.enabled") {
 		// start the connection hub
 		go connectionhub.ConnectionHub.Run()
 		logrus.Infoln("Enabling WebSockets")
--- a/rest/cloudevents/cloudEvents.go
+++ b/rest/cloudevents/cloudEvents.go
@@ -11,7 +11,6 @@ import (
 
 type DataContentType string
 type SpecVersion string
-type Data string
 
 const (
 	ApplicationJson DataContentType = "application/json"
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -57,7 +57,7 @@ func main() {
 
 	// We might want to setup some event listeners at some point, but the pod will
 	// have to restart for these to take effect. We can't enable and disable websockets on the fly
-	if featureflags.IsEnabled("chrome-service.websockets.enabled") {
+	if true || featureflags.IsEnabled("chrome-service.websockets.enabled") {
```
Hyperkid123, line 60: This change should not be included

rest/kafka/consumers.go
```diff
@@ -44,21 +44,32 @@ func startKafkaReader(r *kafka.Reader) {
...
-						Roles:         p.Data.Roles,
-						Organizations: p.Data.Organizations,
-					},
-					Broadcast: p.Data.Broadcast,
-					Data:      data,
-				}
-				if p.Data.Broadcast {
-					logrus.Infoln("Emitting new broadcast message from kafka reader: ", string(newMessage.Data))
-					connectionhub.ConnectionHub.Broadcast <- newMessage
+				//validDct := cloudevents.DataContentType(p.DataContentType)
```
Hyperkid123, line 47: This is probably some left over comment

rest/kafka/consumers.go
```diff
@@ -44,21 +44,32 @@ func startKafkaReader(r *kafka.Reader) {
...
+				err := p.DataContentType.IsValid()
+				if err != nil {
+					log.Println("DCT needs to be json formatted", p, err)
```
Hyperkid123, line 50: We should probably say "kafka message payload" instead of DCT

rest/kafka/consumers.go
```diff
@@ -44,21 +44,32 @@ func startKafkaReader(r *kafka.Reader) {
...
 				} else {
-					logrus.Infoln("Emitting new message from kafka reader: ", string(newMessage.Data))
-					connectionhub.ConnectionHub.Emit <- newMessage
+					err := p.SpecVersion.IsValid()
+					if err != nil {
+						log.Println("Spec version needs to be 1.0.2", p, err)
```
Hyperkid123, line 54: Instead of the number, we should list all possible versions. Making another type specific method like `toString` can be useful here.

Hyperkid123 (commented):
We also have to validate the `source` field.

rest/cloudevents/cloudEvents.go
```diff
@@ -37,6 +38,14 @@ func (sv SpecVersion) IsValid() error {
 	return fmt.Errorf("invalid cloud events spec version, expect %v, got %v", V102, sv)
 }
 
+func SourceIsValid(uri string) error {
```
Hyperkid123, line 41: I would like this to be attached to the data type. You will have create a new one for the source which is currently just a string.

The interface should be similar to the content type and version.

rsun19:
Made requested changes!

rest/kafka/consumers.go
```diff
@@ -44,21 +44,36 @@ func startKafkaReader(r *kafka.Reader) {
...
-				if p.Data.Broadcast {
-					logrus.Infoln("Emitting new broadcast message from kafka reader: ", string(newMessage.Data))
-					connectionhub.ConnectionHub.Broadcast <- newMessage
+				err := p.DataContentType.IsValid()
+				if err != nil {
+					log.Println("Kafka message payload needs to be JSON formatted", err)
 				} else {
-					logrus.Infoln("Emitting new message from kafka reader: ", string(newMessage.Data))
-					connectionhub.ConnectionHub.Emit <- newMessage
+					err := p.SpecVersion.IsValid()
```
Hyperkid123, line 51: I am wondering if we can just exit early from the function rather than keep nesting the code deeper and deeper. Its a bit hard to read.

rsun19:
Extracted `validatePayload` method. Problem with `break`-ing early is that it stops the for loop, and when you run `go run .` again, it caches the previous `go run cmd/kafka/testMessage.go` requests, so the only way to fix it is to restart `make infra`. So instead, the code is only nested one extra layer deep and all other validation methods are now separate for readability.

Hyperkid123 (approved):
:100:

PR #176, opened 2023-06-23
Updating more CGO Enabled references for the security compliance branch

xxlhacker:
There are two more references to CGO enabled that need to update to the value of 1


casey-williams-rh (approved):
👍

PR #186, opened 2023-07-05
[RHCLOUD-26339] add search terms for compliance

BlakeHolifield:


static/beta/prod/navigation/rhel-navigation.json
```diff
@@ -214,6 +214,14 @@
         {
           "title": "Compliance",
           "expandable": true,
+          "alt_title": [
```
Hyperkid123, line 217: I don't think this one is used anywhere in the services.json right? It won't be uploaded to the search index.

Hyperkid123 (changes requested):
You either have to change items that are in the services.json or create a custom index entry

Hyperkid123 (approved):

PR #187, opened 2023-07-05
[RHCLOUD-26343] Enhance Search Results: Malware

rsun19:


static/beta/itless/navigation/rhel-navigation.json
```diff
@@ -161,6 +161,14 @@
               "href": "/insights/malware/systems",
               "product": "Red Hat Insights"
             }
+          ],
+          "alt_title": [
```
Hyperkid123, line 165: This entry is never used directly in the `services.json` files this will never get uploaded to the search index. Either find a different link which is referenced or create a new search entry in the `static-services-entries.json` file.

Hyperkid123 (changes requested):

static/beta/itless/navigation/rhel-navigation.json
rsun19: Let me know if the changes that I've made is fine!

Hyperkid123 (approved):

PR #194, opened 2023-07-10
use credit card icon for services for now

InsaneZein:
Need to temporarily use credit card icon for Subscriptions tile

part of [RHCLOUD-25606](https://issues.redhat.com/browse/RHCLOUD-25606)

### DIFF
--- a/static/beta/stage/services/services.json
+++ b/static/beta/stage/services/services.json
@@ -261,7 +261,7 @@
   },
   {
     "id": "subscription-services",
-    "icon": "ShoppingCartIcon",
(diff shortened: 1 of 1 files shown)


Hyperkid123 (approved):

Hyperkid123:
@InsaneZein you might have to add the icon the to mapper in chrome UI. I can't remember if we already enabled this one.

InsaneZein:
> @InsaneZein you might have to add the icon the to mapper in chrome UI. I can't remember if we already enabled this one.

I think this one is already enabled because it is also used by Spend Management

PR #219, opened 2023-07-17
[RHCLOUD-26973] Enhance Search Results: RHEL Content Repositories

rsun19:


Hyperkid123 (commented):
@rsun19 is there a description defined we can add to these entries? We are currently missing one.

Hyperkid123 (approved):

PR #237, opened 2023-07-26
fix: make kafka optional

SteveHNH:
Wrap the kafka config in an if statement so we can run where we do not have kafka available


casey-williams-rh (approved):
👍

PR #242, opened 2023-07-28
fix subscription rhel route

InsaneZein:


InsaneZein:
this work has already been done

PR #251, opened 2023-08-09
Move subs usage to /subs/usage

karelhala:


karelhala:
ping @cdcabrera

cdcabrera (approved):
We can confirm `acs` and `data-science` will bring up RHACS and RHODS. If ever needed you can also use the full product alias such as `rhacs` and `rhods`

lgtm!

PR #254, opened 2023-08-14
Updates for renamed Seats Admin module

wscalf:
Based on this: https://github.com/RedHatInsights/chrome-service-backend/commit/7a27bc3d2de343bc0921624b1d71c6385fdbea28

### DIFF
--- a/static/stable/stage/modules/fed-modules.json
+++ b/static/stable/stage/modules/fed-modules.json
@@ -970,8 +970,8 @@
           }
       ]
   },
-  "seatsAdmin": {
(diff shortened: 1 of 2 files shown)


Hyperkid123 (approved):
LGTM

Just a note. The `apps/seats-admin-ui/fed-mods.json` manifest file is still not reachable. Assuming the Akamai rule is not yet enabled for your UI container. So the UI will not show up just yet.

Hyperkid123:
Yes that would be it.

PR #256, opened 2023-08-14
Add seats admin prod without navigation

wscalf:


Hyperkid123:
Wait until we have the release date information.

PR #257, opened 2023-08-15
Add segment API keys for ansible modules

apinkert:
For [RHCLOUD-27568](https://issues.redhat.com/browse/RHCLOUD-27568)


Hyperkid123 (changes requested):
We also need the `beta` config files to be updated.

PR #261, opened 2023-08-23
content-sources goes GA

jlsherrill:


Hyperkid123:
cc @radekkaluzik

jlsherrill:
We're shooting for today, alongside provisioning

Hyperkid123 (approved):

jlsherrill:
@Hyperkid123 i can confirm that what want is just to remove the lab icon and redirect to preview here (picture from prod-stable): 
image

What i wasn't 100% sure on, is to how to do that.  I presume removing isBeta:true will

PR #265, opened 2023-08-28
Add service accounts to fed mods config

karelhala:
### Description

Let's add service accounts to prod-preview environment. Since the app is not yet available on Akamai we'll have to hide it in the nav behind `platform.iam.service-accounts` feature flag.


fhlavac (approved):

Hyperkid123 (commented):
@karelhala you added it to prod stable not prod preview.

Hyperkid123 (approved):

PR #273, opened 2023-09-06
Adjust subscription inventory module selection

karelhala:


static/beta/stage/navigation/subscriptions-navigation.json
```diff
@@ -9,8 +9,8 @@
       "href": "/subscriptions/overview"
     },
     {
-      "id": "subscriptionsInventory",
-      "appId": "subscriptionsInventory",
+      "id": "subscriptionInventory",
```
Hyperkid123, line 12: @karelhala the ID is referenced in the services.json file. We have to fix that as well.

karelhala: Let's keep it as is, so we don't have to change everything everywhere.

Hyperkid123 (approved):

PR #280, opened 2023-09-14
Adds Lightspeed nav menu to stage stable and prod beta

wscalf:
Soft-launches the seats admin interface for Red Hatters in preview mode.

This PR is in draft state pending some outstanding work around entitlements that needs to be done prior to launch.

### DIFF
--- a/static/beta/prod/navigation/ansible-navigation.json
+++ b/static/beta/prod/navigation/ansible-navigation.json
@@ -107,6 +107,26 @@
             ]
         },
         {
+            "title": "Ansible Lightspeed with IBM watsonx Code Assistant",
+            "expandable": true,
+            "routes": [
+                {
+                    "id": "seatsAdminUi",
+                    "appId": "ansibleSeatsAdmin",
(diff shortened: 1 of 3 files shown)


static/beta/prod/navigation/ansible-navigation.json
```diff
@@ -106,6 +106,25 @@
                 }
             ]
         },
+        {
+            "title": "Ansible Lightspeed with IBM watsonx Code Assistant",
+            "expandable": true,
+            "routes": [
+                {
+                    "id": "seatsAdminUi",
```
florkbr, line 114: @wscalf can we please add some descriptions to these new routes? Otherwise when users try to search for these new links they will look odd (we will eventually be enforcing descriptions in these navigation files).

wscalf: Alright, descriptions added (which included updating stage beta, which also needed a description, so one more file now)

karelhala (approved):

PR #292, opened 2023-10-06
Update fed-mods and rhel nav for itless env

aneelac22:


static/beta/itless/modules/fed-modules.json
```diff
@@ -1,6 +1,7 @@
 {
     "landing": {
         "manifestLocation": "/apps/landing/fed-mods.json",
+        "fullProfile": false,
```
Hyperkid123, line 4: This argument is deprecated, we use the `ssoScopes` config now with the full list of custom scopes.

Hyperkid123 (approved):

PR #302, opened 2023-10-17
flip feature flag logic for notifications + integrations

ryelo:


static/beta/stage/navigation/settings-navigation.json
```diff
@@ -36,7 +36,7 @@
             "permissions": [
                 {
                     "method": "featureFlag",
-                    "args": ["platform.sources.integrations", false]
+                    "args": ["platform.sources.integrations", true]
```
karelhala, line 39: This will surface both `sources` and `sources-integrations` to navigation. It should be false.

static/beta/stage/navigation/settings-navigation.json
```diff
@@ -125,7 +125,7 @@
...
```
karelhala, line 128: This should be `false` because it would show both `sources-integrations` and `integrations` in the nav resulting in double integrations in nav.

PR #339, opened 2023-11-10
feat: update stable-stage config for idmsvc

avisiedo:
This change will allow us to deploy in ephemeral and shared the current
progress between the different parts and get feedback about issues and
enhancements.

It has been updated the navigation title to fit what current is presented by
the frontend.

HMS-2678

### DIFF
--- a/static/stable/stage/main.yml
+++ b/static/stable/stage/main.yml
@@ -84,6 +84,13 @@ edge:
       - v1
     isBeta: true
 
+idmsvc:
+  title: "Directory and Domain Services"
+  api:
+    versions:
+      - v1
+    isBeta: true
+
 image-builder:
   title: Image Builder
   api:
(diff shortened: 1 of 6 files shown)


static/stable/stage/main.yml
```diff
@@ -84,6 +84,13 @@ edge:
       - v1
     isBeta: true
 
+idmsvc:
+  title: "Directory and Domain Services"
+  api:
+    versions:
+      - v1
+    isBeta: true
```
avisiedo, line 92: Copied from `beta/stage` directory; Would this be `false`?

Hyperkid123: Nope, beta means the app is still in development. You can keep it.

Hyperkid123 (approved):

PR #343, opened 2023-11-13
Trusted Content and Artifact Signer changes

Gregory-Pereira:
Blocked due to inability to test locally, encountering errors with unleash server.

1. changing default trusted-content page to /application-services/trusted-content/overview
2. updating trusted-artifact-signer page

### DIFF
--- a/static/beta/prod/navigation/application-services-navigation.json
+++ b/static/beta/prod/navigation/application-services-navigation.json
@@ -54,15 +54,13 @@
       ]
     },
     {
-      "groupId": "trusted-content",
       "title": "Source Code Security",
-      "icon": "shield",
       "navItems": [
         {
           "id": "trustedContentOverview",
           "appId": "trustedContent",
           "title": "Trusted Content",
-          "href": "/application-services/trusted-content",
+          "href": "/application-services/trusted-content/overview",
           "icon": "TrustedContentIcon",
           "description": "Increase trust and integrity in source code and accelerate the application development process.",
           "alt_title": [
--- a/static/beta/stage/navigation/application-services-navigation.json
+++ b/static/beta/stage/navigation/application-services-navigation.json
@@ -54,15 +54,13 @@
       ]
     },
     {
-      "groupId": "trusted-content",
       "title": "Source Code Security",
-      "icon": "shield",
       "navItems": [
         {
           "id": "trustedContentOverview",
           "appId": "trustedContent",
           "title": "Trusted Content",
-          "href": "/application-services/trusted-content",
+          "href": "/application-services/trusted-content/overview",
           "icon": "TrustedContentIcon",
           "description": "Increase trust and integrity in source code and accelerate the application development process.",
           "alt_title": [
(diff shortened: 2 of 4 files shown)


static/beta/prod/navigation/application-services-navigation.json
```diff
@@ -57,14 +57,14 @@
       "groupId": "trusted-content",
       "title": "Source Code Security",
       "icon": "shield",
```
Hyperkid123, line 59: This attribute should be removed.

static/beta/prod/navigation/application-services-navigation.json
```diff
@@ -57,14 +57,14 @@
       "groupId": "trusted-content",
```
Hyperkid123, line 57: This attribute should be removed.

static/beta/prod/navigation/application-services-navigation.json
```diff
@@ -57,14 +57,14 @@
...
       "title": "Source Code Security",
       "icon": "shield",
-      "navItems": [
+      "routes":[
```
Hyperkid123, line 60: You need to add `"expandable": true` if you want expandable navigation

Hyperkid123 (changes requested):
There are additional changes required for all the nav files.

You are changing the link pathnames. While clicking in the UI, the app does not seem to be adjusted for the new routing.

Gregory-Pereira:
I will try to address these comments. I had issues building the application for whatever reason to test it locally, but I wanted to model it off of the Application Services sub menu (with cloud icon) available here: https://console.redhat.com/application-services/overview, in that the goal is to have 2 menus (overview and trusted artifact signer) under the trusted-content category.

static/beta/prod/navigation/application-services-navigation.json
Gregory-Pereira: We do not want it to be expandable, only to have 2 entries in the `Source code security` menu, one for overview and one for trusted-artifact-signer.

Gregory-Pereira:
So the issue is only from navigating within the trusted-content module, if I navigate from data-science for instance to trusted-artifact-signer directly there are no problems, but when flipping between the two (trusted-content overview and trusted artifact signer) page does not change but the endpoint its calling seems right. Because of this I think that this issue is either due to the `href` for the overview just being for the whole trusted-content directory. 

Was hoping you could expand on why I am removing the `groupId` and `Icon` for trusted-content @Hyperkid123 ?

Hyperkid123:
> Was hoping you could expand on why I am removing the groupId and Icon for trusted-content @Hyperkid123 ?

Because its a different type of navigation item now. You can read more about the schema here: https://github.com/RedHatInsights/chrome-service-backend/blob/main/docs/cloud-services-config.md#making-changes and here: https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/docs/chrome/docs.md#adding-a-new-top-level-link-to-the-chrome-left-navigation

Gregory-Pereira:
Closing this, no longer needed. Issue got fixed.

PR #348, opened 2023-11-20
Cost Management MFE

dlabrecq:
Updated Cost Management microfrontend, so Federated Modules can be consumed in stage and prod envs.

### DIFF
--- a/static/stable/prod/modules/fed-modules.json
+++ b/static/stable/prod/modules/fed-modules.json
@@ -199,6 +199,9 @@
             }
         ]
     },
+    "costManagementMfe": {
+        "manifestLocation": "/apps/cost-management-mfe/fed-mods.json"
+    },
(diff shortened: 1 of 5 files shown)


static/stable/prod/modules/fed-modules.json
```diff
@@ -199,6 +199,9 @@
             }
         ]
     },
+    "costManagementMfe": {
+        "manifestLocation": "/apps/cost-management-mfe/fed-mods.json"
```
Hyperkid123, line 203: This will require an empty modules array. Chrome currently expects the array to exist. We will address it on the UI side, but don't want to block your release.
```
modules: []
```

dlabrecq: Added the empty modules array

Hyperkid123 (approved):

PR #356, opened 2023-12-04
[DO NOT MERGE] debugging user-preferences URL change

florkbr:
@InsaneZein this is what we'll need to do in order to update the user-preferences URL. I've only updated stage preview here.

We would also need this change on top of your `user-preferences-frontend` PR: https://github.com/RedHatInsights/user-preferences-frontend/commit/5097393adc2535322707916fe21032186ec176e3.

PR https://github.com/RedHatInsights/chrome-service-backend/pull/352 will only update the navigation files, not the location of fed-mods.

### DIFF
--- a/static/beta/stage/modules/fed-modules.json
+++ b/static/beta/stage/modules/fed-modules.json
@@ -18,21 +18,21 @@
     },
     "userPreferences": {
         "manifestLocation": "/apps/user-preferences/fed-mods.json",
-        "defaultDocumentTitle": "User Preferences | User Preferences",
+        "defaultDocumentTitle": "My Notification Preferences | User Preferences",
         "isFedramp": true,
         "modules": [
             {
-                "id": "user-preferences-notifications",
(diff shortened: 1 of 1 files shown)


static/beta/stage/modules/fed-modules.json
```diff
@@ -18,21 +18,21 @@
     },
     "userPreferences": {
         "manifestLocation": "/apps/user-preferences/fed-mods.json",
-        "defaultDocumentTitle": "User Preferences | User Preferences",
+        "defaultDocumentTitle": "My Notification Preferences | User Preferences",
```
florkbr, line 21: Actually, looking at https://issues.redhat.com/browse/RHCLOUD-29464 we probably want this to be `Notification Preferences`

InsaneZein: Yup that is correct. The Title is `Notifications Preferences` but the url is `my-notification-preferences`

PR #367, opened 2023-12-19
fixes broken user-pref path

InsaneZein:
depends on https://github.com/RedHatInsights/user-preferences-frontend/pull/179

### DIFF
--- a/static/beta/prod/modules/fed-modules.json
+++ b/static/beta/prod/modules/fed-modules.json
@@ -20,17 +20,17 @@
         "defaultDocumentTitle": "Notification Preferences | User Preferences",
         "modules": [
             {
-                "id": "my-notification-preferences-notifications",
+                "id": "user-preferences-notifications",
                 "module": "./RootApp",
                 "routes": [
                     {
-                        "pathname": "/my-notification-preferences"
+                        "pathname": "/user-preferences"
                     },
                     {
-                        "pathname": "/my-notification-preferences/notifications"
+                        "pathname": "/user-preferences/notifications"
                     },
                     {
-                        "pathname": "/settings/notifications/my-notification-preferences"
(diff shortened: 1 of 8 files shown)


aferd (approved):

Hyperkid123 (changes requested):
Can you elaborate a bit on how exactly are the links broken? The changes in the linked PR do not change any routing. The `appUrl` affects only local development but has no effect on routing in deployed environment.

Also, I don't see any link changes in the navigation files that still reference the `/user-preferences/*` pathnames.

InsaneZein:
@Hyperkid123 right now in stage the user preferences link directs to a blank page, because of my previous attempt to change the user preferences url. This fixes that

static/beta/prod/modules/fed-modules.json
```diff
@@ -17,20 +17,20 @@
...
+        "defaultDocumentTitle": "Notification Preferences | User Preferences",
         "modules": [
             {
-                "id": "user-preferences-notifications",
+                "id": "my-notification-preferences-notifications",
                 "module": "./RootApp",
                 "routes": [
                     {
-                        "pathname": "/user-preferences"
+                        "pathname": "/my-notification-preferences"
```
karelhala, line 27: We should not change the URL as it requires additional changes in Akamai redirects. The SPUR is only about updating the titles.

InsaneZein: The ticket for the SPUR has a comment talking about updating the URL https://issues.redhat.com/browse/RHCLOUD-29464

InsaneZein: I've gone ahead and reverted the URL changes.

Hyperkid123:
@karelhala I'll leave the decision on you :)

karelhala (approved):

PR #379, opened 2024-01-15
Setup debug logs for favoriting servives.

Hyperkid123:
Part of: RHCLOUD-28789


deploy/clowdapp.yml
```diff
@@ -154,6 +159,7 @@ objects:
       SEARCH_CLIENT_SECRET_STAGE: dGVzdFZhbHVl
       INTERCOM_DBAAS: dGVzdFZhbHVl
       INTERCOM_DBAAS_DEV: dGVzdFZhbHVl
+      DEBUG_FAVORITES_ACCOUNT_1: OTk5
```
BlakeHolifield, line 162: This is fine for now but it might be easier to handle multiple as an env var if needed in the future

rest/service/favoritePageService.go
```diff
@@ -49,6 +53,15 @@ func UpdateFavoritePage(favoritePage models.FavoritePage) error {
 	return database.DB.Model(&models.FavoritePage{}).Where("pathname = ?", favoritePage.Pathname).Update("favorite", favoritePage.Favorite).Error
 }
 
+func debugFavoritesEntry(accountId uint, payload models.FavoritePage) {
+	c := config.Get()
+	for _, i := range c.DebugConfig.DebugFavoriteIds {
+		if i == fmt.Sprint(accountId) {
+			logrus.Warningln(fmt.Sprintf("\n_____\nDEBUG_FAVORITES_ACCOUNT_ID: %d\nDEBUG_FAVORITES_PATH: %s\nDEBUG_FAVORITES_FLAG: %s\n_____", accountId, payload.Pathname, fmt.Sprint(payload.Favorite)))
```
BlakeHolifield, line 60: Should this be `logrus.Debugln(...)` ?

BlakeHolifield (approved):

rest/service/favoritePageService.go
Hyperkid123: @BlakeHolifield we would have to set the log level to debug which would undo the effort updating the logs. :(

PR #381, opened 2024-01-16
[RHCLOUD-30109] Update api server to handle batch last-visited updates

BlakeHolifield:
* POST `/last-visited` now updates from local storage and overwrites the previous data
* GET `/last-visited` filters the db metadata

### DIFF
--- a/rest/routes/lastVisited.go
+++ b/rest/routes/lastVisited.go
@@ -2,23 +2,20 @@ package routes
 
 import (
 	"encoding/json"
-	"net/http"
-
 	"github.com/RedHatInsights/chrome-service-backend/rest/models"
 	"github.com/RedHatInsights/chrome-service-backend/rest/service"
 	"github.com/RedHatInsights/chrome-service-backend/rest/util"
 	"github.com/go-chi/chi/v5"
 	"github.com/sirupsen/logrus"
+	"net/http"
 )
 
 func StoreLastVisitedPages(w http.ResponseWriter, r *http.Request) {
 	user := r.Context().Value(util.USER_CTX_KEY).(models.UserIdentity)
 	userId := user.ID
-	var recentPages []models.LastVisitedPage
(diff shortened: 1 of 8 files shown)


rest/routes/lastVisited.go
```diff
@@ -28,19 +25,18 @@ func StoreLastVisitedPage(w http.ResponseWriter, r *http.Request) {
...
-	err = service.HandlePostLastVisitedPages(userId, currentPage)
+	err = service.HandlePostLastVisitedPages(recentPages.Pages, userId)
 	if err != nil {
 		panic(err)
 	}
-
 	pages, err := service.GetUsersLastVisitedPages(userId)
 
 	if err != nil {
 		panic(err)
```
Hyperkid123, line 35: Note for the future, we should standardize response handling including errors. I've started a little bit on the dashboard PR.

rest/service/lastVisitedService.go
```diff
@@ -3,68 +3,36 @@ package service
...
-		// Lock this row until updated
-		err := database.DB.Clauses(clause.Locking{Strength: "UPDATE"}).Order("updated_at desc").Where("user_identity_id = ?", accountId).Find(&pages).Error
-		if err != nil {
+// HandlePostLastVisitedPages inserts the most recent pages from chrome. Once they are added,
+// all older entries are removed from the table.
+func HandlePostLastVisitedPages(recentPages []models.LastVisitedPage, accountId uint) error {
+	var ids []uint
+	for _, v := range recentPages {
+		v.UserIdentityID = accountId
+		if err := database.DB.Create(&v).Error; err != nil {
```
Hyperkid123, line 22: We should probably use the batch update gorm patter/: https://gorm.io/gen/create.html#Batch-Insert

It should generate just a single query and it should be a single transaction.

Hyperkid123 (approved):

PR #398, opened 2024-02-05
Dashboard template API enhancements

florkbr:
Specifically:
* Clean up dashboard-template API doc
* fix variant sm typo
* add support for `static` on layout items
* fix validator errors (`x should be less than or equal to 1 for sm variant`) when trying to reupload the default templates

### DIFF
--- a/docs/dashboard-layouts.md
+++ b/docs/dashboard-layouts.md
@@ -30,7 +30,7 @@ On first template requests, if a user does not have any custom template, new is 
 
 The newly generated template becomes the user default.
 
-Each template will contain 4 variants of the configurations bad on screen sizes.
+Each template will contain 4 variants of the configurations based on various screen sizes.
 - sm
 - md
 - lg
@@ -56,7 +56,7 @@ Each user can have multiple layout templates. They can create and remove layouts
 
 ## Layout storage
 
-We are dealing with a uniform JSON object. There are no queries made with regard to the actual layout. No significant data are the from the user POV that would make sens to look trough.
+We are dealing with a uniform JSON object. There are no queries made with regard to the actual layout. No significant data is stored from the user POV that would make sense to look through.
 
 The layout is stored as a JSON blob in the database. The entity itself will have name that can be used for filtering.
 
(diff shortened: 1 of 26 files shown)


docs/dashboard-layouts.md
```diff
@@ -30,7 +30,7 @@ On first template requests, if a user does not have any custom template, new is
 
 The newly generated template becomes the user default.
 
-Each template will contain 4 variants of the configurations bad on screen sizes.
+Each template will contain 4 variants of the configurations based on various screen sizes.
```
florkbr, line 33: @Hyperkid123 I'm noticing locally it is impossible to `copy` the base-template as the copy API expects an int ID. 

For example: `POST http://localhost:8000/api/chrome-service/v1/dashboard-templates/base-template/copy` results in 400 bad request as `base-template` cannot be converted into an int.

rest/database/db.go
```diff
@@ -58,6 +58,9 @@ func Init() {
 	if !DB.Migrator().HasTable(&models.SelfReport{}) {
 		DB.Migrator().CreateTable(&models.SelfReport{})
 	}
+	if !DB.Migrator().HasTable(&models.DashboardTemplate{}) {
```
florkbr, line 61: Seeing errors using any of our user APIs as this table is not initialized.

docs/dashboard-layouts.md
Hyperkid123: Yeah, you can't copy the base as it will be forked for a user when they request a specific dashboard template and it does not exist yet.

That said, this feature will be required once we allow multiple versions of the same template for a user.

rest/database/db.go
Hyperkid123: Have you run the migration script?

rest/models/DashboardTemplate.go
```diff
@@ -54,23 +54,24 @@ type GridItem struct {
 	Height    int    `json:"h"`
 	MaxHeight int    `json:"maxH"`
 	MinHeight int    `json:"minH"`
+	Static    bool   `json:"static"`
```
florkbr, line 57: FWIW this is part of react-grid-view's Layout type here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-grid-layout/index.d.ts#L66. It is marked optional boolean.

I'm proposing we add it to the server tracked struct as without it the UI could only have (on initial load of session - removing localStorage from the equation):
* All widgets locked by default
* All widgets unlocked by default

However with this struct change we can also have:
* Some widgets locked and some widgets unlocked by default

By default as of now, the service defaults `static` to `false`.

rest/database/db.go
florkbr: I thought so... For now I've removed this change and will try again tomorrow.

florkbr: Yep - I think the migration script failed the first time I ran it. This works as expected without the change 👍

Hyperkid123 (approved):

PR #399, opened 2024-02-06
AAP-20333: Adding Ansible as a Managed Service to the navigation bar

zkayyali812:
Jira: https://issues.redhat.com/browse/AAP-20333

Not to be added to production yet. 

Configures the current list of URL structures for stable/beta.
Currently a draft. 

Please feel free to leave comments/questions, especially if something seems strange below

### DIFF
--- a/static/beta/stage/modules/fed-modules.json
+++ b/static/beta/stage/modules/fed-modules.json
@@ -1098,6 +1098,44 @@
             }
         ]
     },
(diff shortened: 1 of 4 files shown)


tpouyer:
Maybe a better way to do this would be to check the status of the flux kustomize component because they have a status API that will only be healthy if all contained components are healthy according to their published health checks

PR #401, opened 2024-02-09
add overview section to quay navbar

Marcusk19:
Adds overview page to quay.io's navbar

### DIFF
--- a/static/beta/prod/navigation/quay-navigation.json
+++ b/static/beta/prod/navigation/quay-navigation.json
@@ -4,6 +4,26 @@
   "description": "Quay builds, analyzes, distributes your container images.",
   "navItems": [
     {
+      "id": "quayOverview",
+      "appId": "quay-ui-plugin",
+      "title": "Overview",
+      "icon": "QuayIoIcon",
+      "description": "Overview information about quay.io",
+      "href": "/quay/overview"
+      "permissions": [
+        {
+            "method": "apiRequest",
+            "args": [
+                {
+                    "url": "/beta/apps/quay/plugin-manifest.json",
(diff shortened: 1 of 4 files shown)


static/beta/prod/navigation/quay-navigation.json
```diff
@@ -3,6 +3,26 @@
...
+      "title": "Overview",
+      "icon": "QuayIoIcon",
+      "description": "Overview information about quay.io",
+      "href": "/quay/overview"
+      "permissions": [
+        {
+            "method": "apiRequest",
+            "args": [
+                {
+                    "url": "/beta/apps/quay/plugin-manifest.json",
```
Hyperkid123, line 18: Is this still required? Is the Quay plugin still supposed to be hidden from some users? Just curious.

Hyperkid123:
Seems like the JSON is malformed somehow
```
11:04:29 1.123 File /src/chrome-service-backend/static/beta/prod/navigation/quay-navigation.json
11:04:29 1.125 panic: invalid character '"' after object key:value pair
```

static/beta/prod/navigation/quay-navigation.json
Marcusk19: Not sure, I just figured it's best to leave it the same as the other navs.

Hyperkid123: @Marcusk19 can I request a follow-up to this? I vaguely remember this was done during the initial release to ensure the nav items show when the deployment is ready. The quay plugins are now in production so this API check will always be true. Which makes the API calls obsolete and wastes resources.

Hyperkid123 (approved):

PR #405, opened 2024-02-14
remove permission check from quay navs

Marcusk19:
permission checks not required for quay anymore


Hyperkid123 (approved):

Hyperkid123:
The JSON is malformed somewhere:
```
16:03:30 1.173 File /src/chrome-service-backend/static/beta/prod/navigation/quay-navigation.json
16:03:30 1.175 panic: invalid character '}' looking for beginning of object key string
```

PR #407, opened 2024-02-21
Remove seatsAdminUi nav items

wscalf:
Removes nav menu items for seatsAdminUi in prod preview, prod stable, stage preview, and stage stable, as it's been deprecated in favor of the new Admin Dashboard.

### DIFF
--- a/static/beta/prod/navigation/ansible-navigation.json
+++ b/static/beta/prod/navigation/ansible-navigation.json
@@ -112,14 +112,6 @@
             "expandable": true,
             "routes": [
                 {
-                    "id": "seatsAdminUi",
-                    "appId": "ansibleSeatsAdmin",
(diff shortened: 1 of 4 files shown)


Hyperkid123 (commented):
The UI is still going to be accessible via direct URL unless you remove the config from `fed-modules.json` files.

Is that OK?

Also, is there a specific date you are targetting? Or can we deploy it whenever we get the chance?

wscalf:
The fed-modules.json is a good point- we have a separate work item to decomission that, but that creates a dependency where this needs to go in first, or it'll leave a dead link. If it gets deployed pretty quickly, that's not a big deal, but it might simplify things for everyone if we do them together.

edewit (approved):

Hyperkid123 (approved):

PR #423, opened 2024-03-07
Show the Configure Events page for non org admin users (Stage)

CodyWMitchell:
[RHCLOUD-30431](https://issues.redhat.com/browse/RHCLOUD-30431)
- Allow users with `integrations:*:*` , `integrations:endpoints:write` to view the Configure Events page in stage beta and preview

### DIFF
--- a/static/beta/stage/navigation/settings-navigation.json
+++ b/static/beta/stage/navigation/settings-navigation.json
@@ -71,7 +71,7 @@
                         "method": "loosePermissions",
                           "args": [
                             "isOrgAdmin",
-                            "integrations::",
+                            "integrations:*:*",
                             "integrations:endpoints:write"
                         ]
                       }
--- a/static/beta/stage/navigation/settings-navigation.json
+++ b/static/beta/stage/navigation/settings-navigation.json
@@ -68,7 +68,12 @@
                     "href": "/settings/notifications/configure-events",
                     "permissions": [
                       {
-                        "method": "isOrgAdmin"
+                        "method": "loosePermissions",
+                          "args": [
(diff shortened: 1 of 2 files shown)


static/stable/stage/navigation/settings-navigation.json
```diff
@@ -54,7 +54,12 @@
                   "href": "/settings/notifications/configure-events",
                   "permissions": [
                     {
-                      "method": "isOrgAdmin"
+                        "method": "loosePermissions",
+                        "args": [
+                            "isOrgAdmin",
```
karelhala, line 59: Same as above.
```suggestion
```

static/beta/stage/navigation/settings-navigation.json
```diff
@@ -68,7 +68,12 @@
                     "href": "/settings/notifications/configure-events",
                     "permissions": [
                       {
-                        "method": "isOrgAdmin"
+                        "method": "loosePermissions",
+                          "args": [
+                            "isOrgAdmin",
```
karelhala, line 73: This won't work, the `isOrgAdmin` should be another permission method. @Hyperkid123 do we have `or` between permission methods or `and`?
```suggestion
```

Hyperkid123: We only have `and`

The permission `isOrgAdmin` does not exist. However, org admin should have the specific permission anyway. So finding using permission strings should work. I might be wrong.

karelhala (approved):
Let's try this approach, org admins should have either integrations all or write permissions anyway.

PR #443, opened 2024-03-25
Use existing widget in landing page layout

Hyperkid123:
https://issues.redhat.com/browse/RHCLOUD-31551

### Changes
- add more strict typing to the layout JSON blob
- add migration to remove non existing widgets from layouts
- update landing page layout to use existing widgets

### DIFF
--- a/cmd/migrate/migrate.go
+++ b/cmd/migrate/migrate.go
@@ -29,9 +29,9 @@ func getValidGridItems(items []models.GridItem) []models.GridItem {
 	return newGridItems
 }
 
-func migrateDashboardWidgets() *gorm.DB {
+func migrateDashboardWidgets(tx *gorm.DB) *gorm.DB {
 	var templates []models.DashboardTemplate
-	res := database.DB.Find(&templates)
+	res := tx.Find(&templates)
 	if res.Error != nil {
 		return res
 	}
@@ -41,7 +41,7 @@ func migrateDashboardWidgets() *gorm.DB {
 		t.TemplateConfig.Lg = datatypes.NewJSONType(getValidGridItems(t.TemplateConfig.Lg.Data()))
 		t.TemplateConfig.Md = datatypes.NewJSONType(getValidGridItems(t.TemplateConfig.Md.Data()))
 		t.TemplateConfig.Sm = datatypes.NewJSONType(getValidGridItems(t.TemplateConfig.Sm.Data()))
-		res = database.DB.Save(&t)
+		res = tx.Save(&t)
 
 		if res.Error != nil {
 			return res
@@ -108,7 +108,7 @@ func main() {
 		panic(bundleRes.Error)
 	}
 
-	dashboardRes = migrateDashboardWidgets()
+	dashboardRes = migrateDashboardWidgets(tx)
 	if dashboardRes.Error != nil {
 		logrus.Error("Unable to migrate database!")
 		tx.Rollback()
--- a/cmd/migrate/migrate.go
+++ b/cmd/migrate/migrate.go
@@ -1,12 +1,55 @@
 package main
 
 import (
+	"strings"
+
 	"github.com/RedHatInsights/chrome-service-backend/rest/database"
 	"github.com/RedHatInsights/chrome-service-backend/rest/models"
 	"github.com/joho/godotenv"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


cmd/migrate/migrate.go
```diff
@@ -1,19 +1,63 @@
...
+	newGridItems := []models.GridItem{}
+	for _, item := range items {
+		widgetTypes := strings.Split(item.ID, "#")
+		if len(widgetTypes) > 0 {
+			widgetType := widgetTypes[0]
+			err := models.AvailableWidgets(widgetType).IsValid()
+			if err == nil {
+				newGridItems = append(newGridItems, item)
+			} else {
+				logrus.Infof("Removing invalid widget type: %s", widgetType)
```
florkbr, line 24: @Hyperkid123 this looks good for now - though we may need to put some more thought into this long term. Such as handling situations where a widget type is renamed.

With this change all of the old widgets would be removed from the dashboards. 

I'm not implying that would happen often or ever (as the widgetType is just an identifier) - just that we should make communications clear in our readme about how those possible situations would be handled (or discouraging them from happening).

rest/models/DashboardTemplate.go
```diff
@@ -184,8 +195,10 @@ type BaseDashboardTemplate struct {
 type BaseTemplates map[AvailableTemplates]BaseDashboardTemplate
 
 type ModuleFederationMetadata struct {
-	Scope      string `json:"scope"`
-	Module     string `json:"module"`
-	ImportName string `json:"importName,omitempty"`
+	Scope      string               `json:"scope"`
+	Module     string               `json:"module"`
+	ImportName string               `json:"importName,omitempty"`
+	Defaults   BaseWidgetDimensions `json:"defaults"`
```
florkbr, line 201: As a follow up can we add some documentation to the readme here https://github.com/RedHatInsights/chrome-service-backend/blob/main/docs/dashboard-layouts.md#widgets-mapping to include `Defaults`?

florkbr:
I tested these changes locally (after setting up my DB to have some old widget flavors). They are properly removed:
Screenshot 2024-03-25 at 3 50 03 PM

I also see the new default metadata coming down for each widget.

One issue I did notice - since the default `landingPage` layout does not have any widgets for the `sm`, `md`, `lg`, or `xl` variants - we run into a UI bug where you cannot drag and drop new widgets as the grid's width/height is 0. That should go away once we implement the grid empty state in the UI. I'll create a ticket for it.

florkbr:
@Hyperkid123 I didn't have a ton of time to debug, but using the migrator (versus accessing the DB directly) resolves the issue:
```diff
diff --git a/cmd/migrate/migrate.go b/cmd/migrate/migrate.go
index 177b749..08d94f2 100644
--- a/cmd/migrate/migrate.go
+++ b/cmd/migrate/migrate.go
@@ -29,9 +29,9 @@ func getValidGridItems(items []models.GridItem) []models.GridItem {
        return newGridItems
 }
 
-func migrateDashboardWidgets() *gorm.DB {
+func migrateDashboardWidgets(tx *gorm.DB) *gorm.DB {
        var templates []models.DashboardTemplate
-       res := database.DB.Find(&templates)
+       res := tx.Find(&templates)
        if res.Error != nil {
                return res
        }
@@ -41,7 +41,7 @@ func migrateDashboardWidgets() *gorm.DB {
                t.TemplateConfig.Lg = datatypes.NewJSONType(getValidGridItems(t.TemplateConfig.Lg.Data()))
                t.TemplateConfig.Md = datatypes.NewJSONType(getValidGridItems(t.TemplateConfig.Md.Data()))
                t.TemplateConfig.Sm = datatypes.NewJSONType(getValidGridItems(t.TemplateConfig.Sm.Data()))
-               res = database.DB.Save(&t)
+               res = tx.Save(&t)
 
                if res.Error != nil {
                        return res
@@ -108,7 +108,7 @@ func main() {
                panic(bundleRes.Error)
        }
 
-       dashboardRes = migrateDashboardWidgets()
+       dashboardRes = migrateDashboardWidgets(tx)
        if dashboardRes.Error != nil {
                logrus.Error("Unable to migrate database!")
                tx.Rollback()
(END)

```

florkbr (approved):

Hyperkid123:
Oh yeah I forgot the table might not exist LOL

PR #445, opened 2024-03-25
Adding virtual-assistant to prod-stable

josejulio:


florkbr:
Context: we are trying to enable VA in prod (with the unleash flag). When turned it on scalprum threw out an error (likely due to the fed mod not being defined for prod-stable). It's already present in the prod-beta fed-mods.

florkbr:
:shipit:

PR #468, opened 2024-04-08
RHCLOUD-31837: Add support for widget permissions via widget mapping read only endpoint

florkbr:
We need to hide certain widgets from a users dashboard as well as from the widget drawer when specific permission criteria are met.

I've modeled these new structures based on what we have today in the chrome backend navigation schema [here](https://github.com/RedHatInsights/chrome-service-backend/blob/main/static/navigationSchema.json#L5-L21)

See https://issues.redhat.com/browse/RHCLOUD-31837


rest/models/DashboardTemplate.go
```diff
@@ -246,9 +245,9 @@ const (
 	OpenShiftIcon        WidgetIcons = "OpenShiftIcon"
 	EdgeIcon             WidgetIcons = "EdgeIcon"
 	AnsibleIcon          WidgetIcons = "AnsibleIcon"
-	QuayIcon			 WidgetIcons = "QuayIcon"
```
florkbr, line 249: This was a spacing issue upstream :(

Hyperkid123 (approved):

PR #474, opened 2024-04-11
 define widgetLayout in fed-modules for prod beta

epwinchell:


static/beta/prod/modules/fed-modules.json
```diff
@@ -1004,6 +1004,20 @@
...
     },
+    "widgetLayout": {
+        "manifestLocation": "/apps/widget-layout/fed-mods.json",
+        "modules": [
+          {
+              "id": "widgetLayout",
+              "module": "./RootApp",
+              "routes": [
+                  {
+                      "pathname": "/staging/widget-layout"
```
florkbr, line 1015: FWIW this is only for our test page of the widget-layout. We should probably remove this and copy what VA is doing (`"modules": []`) once we promote to stable/released. 

I'm also not convinced we updated routes in akamai in prod to enable this path - so it may not work anyway. 

For now looks good to me though to unblock stage-preview.

florkbr (approved):

PR #475, opened 2024-04-15
feat: updates alt_title keywords for Tasks in HCC search

petrblaho:
This is needed to overcome some shortcomings of current search solution,
mainly case sensitivnes.

RHINENG-8428

### DIFF
--- a/static/beta/itless/navigation/rhel-navigation.json
+++ b/static/beta/itless/navigation/rhel-navigation.json
@@ -413,14 +413,23 @@
           "subtitle": "Red Hat Insights for RHEL",
           "description": "Perform Red Hat-recommended analysis across your Red Hat Enterprise Linux fleet, like upgrade and conversion readiness.",
           "alt_title": [
-              "Tasks",
-              "task",
-              "CentOS conversion",
-              "Pre Conversion analysis",
-              "Pre-Conversion analysis",
-              "Convert to RHEL",
-              "C2R",
-              "CentOS"
+            "Tasks",
+            "tasks",
+            "task",
+            "CentOS Conversion",
+            "centos conversion",
+            "Pre Conversion Analysis",
+            "pre conversion analysis",
+            "Pre-Conversion Analysis",
+            "pre-conversion analysis",
+            "Convert to RHEL",
+            "convert to rhel",
+            "C2R",
+            "c2r",
+            "CentOS",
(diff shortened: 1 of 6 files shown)


static/beta/itless/navigation/rhel-navigation.json
```diff
@@ -413,14 +413,20 @@
...
-              "Tasks",
-              "task",
-              "CentOS conversion",
-              "Pre Conversion analysis",
-              "Pre-Conversion analysis",
-              "Convert to RHEL",
-              "C2R",
-              "CentOS"
+            "Tasks",
+            "tasks""task",
```
gkarat, line 417:
```suggestion
            "tasks", "task",
```

gkarat: And I see many other places with the same issue (won't be commenting on them)

static/beta/stage/navigation/rhel-navigation.json
```diff
@@ -541,14 +541,22 @@
...
-              "CentOS conversion",
-              "Pre Conversion analysis",
-              "Pre-Conversion analysis",
-              "Convert to RHEL",
-              "C2R",
-              "CentOS"
+            "Tasks",
+            "tasks"
+            "task",
+            "CentOS conversion",
```
gkarat, line 547: Can we also add "centos conversion" and "conversion"? Since the search service sometimes also doesn't work properly with partial matches

static/beta/itless/navigation/rhel-navigation.json
petrblaho: yup, my slopy copy-pasta

static/beta/stage/navigation/rhel-navigation.json
petrblaho: it also doesn't work correctly with multi-word matches... even now "CentOS conversion" will not return anything. "conversion" will return Tasks correctly. I will add capitalized version of this string. "centos" and its variants are few lines lower.

gkarat (approved):
Awesome! Thanks @petrblaho

gkarat:
@Hyperkid123, could we also have your brief look/help us merge the PR?

Hyperkid123 (approved):

petrblaho:
thank you guys

PR #477, opened 2024-04-15
add config for subscriptions widget

CodyWMitchell:
### **Merge only after https://github.com/RedHatInsights/subscription-inventory-ui/pull/171 has been merged**

[RHCLOUD-31202](https://issues.redhat.com/browse/RHCLOUD-31202)
- Adds config to support subscriptions widget being added to landing page
- widget hidden behind feature flag `chrome-service.subscriptions-widget.enabled`


Hyperkid123:
@CodyWMitchell you can hide the new widget behind a FF. Chrome service is already hooked to unleash. An example is here: https://github.com/RedHatInsights/chrome-service-backend/blob/16b99e76ec35bddccf395c6fa431572eba0e2985/main.go#L72

To add a new FF to the unleash instance and you should be good to go.

You want to adjust the default layout and also the widget mapping. The mapping because we don't want the new widgets show in the add widget panel and the default layout because we don't want to show the widget in new layouts.

rest/service/dashboardTemplate.go
```diff
@@ -160,6 +160,19 @@ var (
 				Title: "Advanced Cluster Security",
 			},
 		},
+		models.Subscriptions: models.ModuleFederationMetadata{
+			Scope:    "landing",
```
karelhala, line 164: The scope is connected to the appName's module scope. So for the subs inventory it should be `subscriptionInventory`
```suggestion
			Scope:    "subscriptionInventory",
```

Hyperkid123:
@karelhala @CodyWMitchell these commits need to be squashed in the future. It's going to be a pain to revert if for some reason we would have to.

PR #484, opened 2024-04-19
update dockerfile COPY command

aneelac22:


Dockerfile
```diff
@@ -1,6 +1,13 @@
...
 WORKDIR $GOPATH/src/chrome-service-backend/
-COPY . .
+COPY api /api
+COPY cmd /cmd
+COPY config /config
+COPY rest /rest
+COPY static /static
+COPY go.mod /go.mod
+COPY go.sum /go.sum
+COPY main.go /main.go
```
Hyperkid123, line 10: We also ant the `spec` dir

PR #500, opened 2024-05-09
fix(RHINENG-7102): Fix missing search entries for Insights

gkarat:
Fixes https://issues.redhat.com/browse/RHINENG-7102.

This is another attempt to put all the Insights entries to shape.

This mainly addresses the following issues:

- Searching for Advisor should give me all Advisor apps (RHEL, OCP, Ansible).
- Searching for Patch should also return Patch systems and Patch packages pages.
- Searching for policy names (PCI, HIPAA, DSS, CSI, C2S) should return Compliance app
- Searching for ‘rules’ should give me Advisor recommendations (due to former name ‘Insights rules’) and Compliance (since it’s still using rules)
- searching for incident gives me nothing, I should be presented with Advisor (‘incidents’ work however)
- Searching for 'groups', one of the resutls should be Inventory groups

### DIFF
--- a/cmd/search/static-services-entries.json
+++ b/cmd/search/static-services-entries.json
@@ -477,7 +477,8 @@
           "machine(s)",
           "server(s)"
         ]
-      }
+      },
+      "rhel.groups"
     ]
   },
   {


gkarat:
@Hyperkid123, could you or someone from the platform take a look? The PR is split into more commits so that's easier to verify

Hyperkid123 (approved):

PR #518, opened 2024-05-22
Remove unused business-services bundle

dlabrecq:
Removed the unused business-services bundle. The HCS UI was moved to the subscriptions bundle a while ago -- no need to maintain this.


Hyperkid123 (commented):
Some references to this bundle are within the chrome UI code. EG: https://github.com/RedHatInsights/insights-chrome/blob/f138e5614fb857e896e8406fe8edf4206f501806/src/components/AppFilter/useAppFilter.ts#L26

Needs to be removed first to prevent unexpected errors.

dlabrecq:
Thanks. I created a PR to remove the business-services filter
https://github.com/RedHatInsights/insights-chrome/pull/2843

Hyperkid123 (approved):

PR #541, opened 2024-06-26
fixing iam groups nav

Fewwy:
Apparently, I added a bit too many things in unnecessary places. This PR fixes that

### DIFF
--- a/static/beta/itless/services/services.json
+++ b/static/beta/itless/services/services.json
@@ -24,7 +24,8 @@
         "title": "Red Hat Enterprise Linux",
         "links": [
           "rhel.inventory",
-          "rhel.groups"
+          "rhel.groups",
+          "rhel.workspaces"
         ]
       },
       {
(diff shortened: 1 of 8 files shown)


static/stable/stage/services/services.json
```diff
@@ -179,8 +178,7 @@
         "links": [
           "rhel.imageBuilder",
           "rhel.inventory",
-          "rhel.groups",
-          "rhel.workspaces"
+          "rhel.groups"
```
gkarat, line 181: Why we are removing "rhel.workspaces"?

static/beta/itless/services/services.json
```diff
@@ -25,8 +24,7 @@
         "title": "Red Hat Enterprise Linux",
         "links": [
           "rhel.inventory",
-          "rhel.groups",
-          "rhel.workspaces"
+          "rhel.groups"
```
gkarat, line 27: Same question here

Fewwy: Because its in the "title": "Identity and Access Management",

Hyperkid123 (approved):

PR #542, opened 2024-06-27
prod workspace nav update

Fewwy:


Hyperkid123 (approved):

static/beta/prod/navigation/rhel-navigation.json
```diff
@@ -39,7 +39,33 @@
...
+          "title": "Workspaces",
+          "href": "/insights/inventory/workspaces",
+          "product": "Red Hat Insights",
+          "description": "Group your Red Hat Enterprise Linux systems for more granular User Access.",
+          "alt_title": [
+            "workspace",
+            "workspaces",
+            "Workspace",
+            "Workspaces"
+          ],
```
gkarat, line 62: nit-pick: this would be nice to preserve "group"-like alternative titles so that users can still be suggested with Workspaces even though they searched for outdated "groups" keyword

gkarat (approved):
LGTM, thanks @Fewwy. Only one minor thing, but not blocking the merge much

PR #572, opened 2024-07-16
Add learning resources creator to beta staging

randomnetcat:
An attempt to expose the application added by https://github.com/RedHatInsights/learning-resources/pull/54.

I'm not sure how exactly I should be doing this, so this is probably wrong?


JetyAdam (approved):

JetyAdam:
It's indeed correct.

PR #583, opened 2024-07-30
Add config for stage activation keys

jlgre:
This adds the config to use activation-keys-ui's deployment instead of sed-frontend. This is part of the effort to decouple those two applications.

Navigation should not have to change I think - as the routes are staying the same. Just the location of the code is changing.

### DIFF
--- a/static/beta/stage/modules/fed-modules.json
+++ b/static/beta/stage/modules/fed-modules.json
@@ -519,7 +519,7 @@
                 "id": "activation-keys",
                 "module": "./RootApp",
                 "routes": [
-                    {"pathname": "/insights/connector/activation-keys"},
+                    {"pathname": "/insights/connector/activation-keys"}
                 ]
             }
         ]
--- a/static/beta/stage/modules/fed-modules.json
+++ b/static/beta/stage/modules/fed-modules.json
@@ -512,6 +512,18 @@
             }
         ]
     },
+    "activationKeys": {
+        "manifestLocation": "/apps/activation-keys/fed-mods.json",
+        "modules": [
+            {
+                "id": "activation-keys",
+                "module": "./RootApp",
+                "routes": [
+                    {"pathname": "/insights/connector/activation-keys"},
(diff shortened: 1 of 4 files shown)


jlgre:
@Hyperkid123 do you mind taking a look at this when you have a minute?

Hyperkid123 (commented):
The config looks good. I assume the JS assets are not available yet? We can merge this as a prep as it won't really change much until the assets are accessible.

jlgre:
> The config looks good. I assume the JS assets are not available yet? We can merge this as a prep as it won't really change much until the assets are accessible.

The js assets should be available - I have the app deploying to stage already via app-interface. The old app is also still deploying so I'm not totally sure how they will work when they are both deployed, but I have a PR approved and ready to merge to remove the routes for this app so they no longer conflict.

Hyperkid123:
@jlgre the URL currently throws 404. Have you requested Akamai routing changes yet?

jlgre:
I have not - I didn't see that step in the docs. Do you mind pointing me to where the information about that is?

justinorringer:
Hello! here's the doc on it

https://inscope.corp.redhat.com/docs/default/Component/consoledot-pages/containerized-frontends/deploy-stage-and-prod/#akamai-configuration

But really, just clone https://issues.redhat.com/browse/RHCLOUD-25708 for stage, and we will take care of it :)

Here's the one for prod too, for when you have it deployed there https://issues.redhat.com/browse/RHCLOUD-25710

jlgre:
@justinorringer thank you! Cloned!

jlgre:
The assets are in stage now so I believe this should be ready to go - I see the Konflux build is failing but am not sure that has anything to do with my new changes

Hyperkid123:
@jlgre the konflux failure is fine, we have switched konflux namespace while this PR was opened so I think its just confused.

Hyperkid123 (approved):

PR #585, opened 2024-08-02
Adding ssoScopes to virtual assistant

tahmidefaz:


static/beta/stage/modules/fed-modules.json
```diff
@@ -1151,6 +1151,11 @@
     },
     "virtualAssistant": {
         "manifestLocation": "/apps/virtual-assistant/fed-mods.json",
+        "config": {
+            "ssoScopes": [
+                "api.iam.service_accounts"
+            ]
+        },
```
tahmidefaz, line 1158: If we add this `ssoScope`, does it mean people who don't have service_accounts access will be blocked from virtual assistant? If so, please close this PR, since it'd mean that people without service accounts access will be blocked from accessing the landing page.

josejulio: It is worth to note that this component gets loaded in the landing page as a federated module. 
We need this scope in the token as part of one flow. 

Please let us know If there is a better way to use this scope once we are sure it is required (I.e. when the user wants to create a service account).

tahmidefaz:
Closing, as per discussion over slack.

PR #597, opened 2024-08-12
Enable Learning Resources for IT-less

casey-williams-rh:


InsaneZein:
I noticed this isn't in static/beta/itless. Do you want to add it there as well?

casey-williams-rh:
@InsaneZein , That is a good catch. I'll add it

InsaneZein (approved):

PR #628, opened 2024-08-16
fix: Remove Groups entry from stage/stable navigation

gkarat:
Quick PR to remove the Groups entry completely from the stage stable navigation. After the recent updates, it led to a bug where both Workspaces and Groups have become visible.

### DIFF
--- a/static/stable/stage/navigation/rhel-navigation.json
+++ b/static/stable/stage/navigation/rhel-navigation.json
@@ -27,20 +27,6 @@
           "subtitle": "Red Hat Insights for RHEL"
         },
         {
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@gkarat 

Seems like the services json is not valid
```
subprocess exited with status 2
subprocess exited with status 2
Error: building at STEP "RUN make parse-services": exit status 2
```

gkarat:
@Hyperkid123, could you please take a look once more? It's quite annoying thingy on stage and we better remove it ASAP 😄

Hyperkid123 (approved):

PR #663, opened 2024-09-05
Removes Content Repositories

shannon-donahue:
Removes content repositories page from UI as well as its nav reference in itless.


florkbr (approved):
The content sources API is not currently available in our itless envs - so looks good to me. If we eventually add them - we can revert this PR! Thanks @shannon-donahue!

PR #664, opened 2024-09-09
Disable WS broadcast.

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-33399


Hyperkid123:
@florkbr nope it is not used

PR #674, opened 2024-09-19
HMS-4711: make idmsvc visible only in preview mode

pvoborni:
https://issues.redhat.com/browse/HMS-4711

### DIFF
--- a/static/stable/stage/navigation/settings-navigation.json
+++ b/static/stable/stage/navigation/settings-navigation.json
@@ -97,6 +97,11 @@
             "description": "Register identity and access systems to enable machines to automatically join a domain",
             "href": "/settings/idmsvc",
             "icon": "PlaceholderIcon",
+            "permissions": [
+              {
+                "method": "isBeta"
+              }
+            ],
             "alt_title": [
                 "directory",
                 "domain",


frasertweedale (approved):
Petr recorded a demo to show the effect of the change (https://drive.google.com/file/d/1POwqGz8QBEXSoeraPCCZyYboXsEMMp19/view ; Red Hat internal).  Everything LGTM.

florkbr:
@pvoborni @frasertweedale a few things I want to point out:
1. After watching the attached video - I want to make it clear this change will only hide the left navigation item when preview is disabled. The behavior of hiding the `All Services` dropdown entry and the favorites item is another change we must already have delivered.
2. These changes work for me trying things out locally (the left nav item hides when I disable preview and refresh the page). **However one thing to note** - since the URL is already in the address bar when looking at the `Directory and Domain Services` UI - after I disable preview and refresh the page, I still have access to the UI - just the left nav item is removed (see screenshot below).

If you also want to disable access to the page via the URL, have a look at our chrome preview context/feature flags: https://github.com/RedHatInsights/insights-chrome/blob/master/docs/preview.md#feature-flags

image

florkbr (approved):
Comments are not blocking so :shipit:

PR #692, opened 2024-10-01
RHCLOUD-35156: Add support to chrome service for active workspace on user obj

florkbr:


cmd/migrate/migrate.go
```diff
@@ -123,10 +125,18 @@ func main() {
 
 	fmt.Println("Seed default value to last visited pages")
 	visitedRes = tx.Model(&models.UserIdentity{}).Where("last_visited_pages IS NULL").Update("last_visited_pages", []byte(`[]`))
-	if bundleRes.Error != nil {
-		fmt.Println("Unable to migrate database!", bundleRes.Error.Error())
+	if visitedRes.Error != nil {
```
florkbr, line 128: I think we had a typo checking the wrong migrator response here (copy pasta)

rest/models/UserIdentity.go
```diff
@@ -29,6 +29,7 @@ type UserIdentity struct {
 	DashboardTemplates []DashboardTemplate               `json:"dashboardTemplates,omitempty"`
 	UIPreview          bool                              `json:"uiPreview"`
 	UIPreviewSeen      bool                              `json:"uiPreviewSeen"`
+	ActiveWorkspace    datatypes.JSONType[Workspace]     `json:"activeWorkspace"`
```
Hyperkid123, line 32: I think this just has to be ID of the workspace. I think it should be a uuid so string for us

florkbr: Updated to use string type over JSON

cmd/migrate/migrate.go
```diff
@@ -144,5 +154,10 @@ func main() {
 	if visitedRes.RowsAffected > 0 {
 		logrus.Infof("Migrated %d user identity visited rows", visitedRes.RowsAffected)
 	}
+
+	if activeWorkspaceRes.RowsAffected > 0 {
+		logrus.Infof("Migrated %d user identity visited rows", activeWorkspaceRes.RowsAffected)
```
florkbr, line 159: FYI I have confirmed this works when migrating my DB locally (started chrome on main, create a user in the DB, and then checked out this PR). I see this from the migrate script:
```bash
2024/10/02 18:33:40 /Users/bflorkie/Development/github/RedHatInsights/chrome-service-backend/cmd/migrate/migrate.go:135
[0.442ms] [rows:1] UPDATE "user_identities" SET "active_workspace"='foobar',"updated_at"='2024-10-02 18:33:40.729' WHERE active_workspace IS NULL AND "user_identities"."deleted_at" IS NULL
INFO[0000] Migrated 1 user identity visited rows        
INFO[0000] Migration complete 
```

And then this from the rest endpoint after (note default workspace is "foobar"):
```bash
➜  chrome-service-backend git:(main) curl -H "x-rh-identity:<token>" http://localhost:8000/api/chrome-service/v1/user

{"data":{"createdAt":"0001-01-01T00:00:00Z","accountId":"12","firstLogin":true,  ... "uiPreviewSeen":false,"activeWorkspace":"foobar"}}
```

rest/service/identity.go
```diff
@@ -115,6 +115,7 @@ func CreateIdentity(userId string, skipCache bool) (models.UserIdentity, error)
 		FavoritePages:    []models.FavoritePage{},
 		SelfReport:       models.SelfReport{},
 		VisitedBundles:   nil,
+		ActiveWorkspace:  "foobar",
```
florkbr, line 118: @karelhala @fhlavac do you know what the default workspace ID should be for all users? I will update this PR accordingly. 

This ID will be used to seed existing users in our DB as well as any new users that are created. The workspace can then be changed later through the `update-active-workspace` rest endpoint.

Hyperkid123 (approved):

PR #694, opened 2024-10-02
Add unleash compose files compatible with podman-compose

catastrophe-brandon:
In a nutshell, there are a variety of issues with the current unleash-compose yaml that prevent it from working with podman-compose. This adds:

1. A compose file that will work with podman
2. A SQL init file that will grant the appropriate privs to unleash_user to run the necessary migrations


Hyperkid123:
@catastrophe-brandon we will probably also need some env variables for the WebPort and Metrics ports configurations. The defaults can be occupied and my system is dying on that with these changes.

Hyperkid123 (approved):

PR #718, opened 2024-10-31
AAP-27533 Add Top Level Cluster List Navigation to CRC 

prat98:
This PR is to help set up a dev workflow for all work in AAP-27533. Using this PR, one can create a local chrome service backend with updated nav components for cluster list. In other apps such as analytics, requests can be proxied to this backend to get the updated top level nav when doing development within a specific app.


karelhala:
Closing because of inactivity, if this is still required please open a new PR.

PR #719, opened 2024-11-04
Enable consuming generating fed-modules.json from FEO env variables

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-36064


Hyperkid123:
Pipeline fix is here: https://github.com/RedHatInsights/chrome-service-backend/pull/720

PR #724, opened 2024-11-11
Disable the Red Hat user only filter we had in place before GA

chrisahl:
Our offering is about to be GA so we need to remove the check that only let Red Hat users access it.

For issue https://issues.redhat.com/browse/AAP-33834


florkbr:
Vulnerability scan failures are unrelated to this PR. The services team will follow up on that separately.

florkbr (approved):

PR #756, opened 2025-01-16
task: remove patchTemplates from itless envs

SteveHNH:
These templates are unneeded in the ITless environment so we do not want the link to appear in the navigation.

### DIFF
--- a/static/beta/itless/navigation/rhel-navigation.json
+++ b/static/beta/itless/navigation/rhel-navigation.json
@@ -144,13 +144,6 @@
                 "systems",
                 "upgrade"
               ]
-            },
-            {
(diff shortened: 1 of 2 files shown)


shannon-donahue (approved):
LGTM

static/beta/itless/navigation/rhel-navigation.json
florkbr, line 1: @SteveHNH just an FYI for the future - these beta files are no longer used (so we shouldn't need to update them any longer). We will remove them in the future.

florkbr:
Holding merge... for some reason Konflux did not run for this PR:
image

PR #758, opened 2025-01-16
RHCLOUD-37348: stop symlinking bundle nav for FEO enhancements

florkbr:


Hyperkid123 (commented):
@florkbr github does not allow me to comment in specific lines...

However the IDs of the nav files should match the pathname partial (for now) so for insights it should be `insights` but its currently `rhel`. I'd also check the application pipeline.

Hyperkid123 (approved):

Hyperkid123 (approved):

PR #775, opened 2025-01-31
Removed Spend Group for itless env

wcater803:


static/beta/itless/services/services.json
florkbr, line 1: @wcater803 just an FYI for the future - the `beta` files are no longer used and don't need to be updated. We can leave this PR as it

florkbr:
Unit tests are failing in Konflux due to: `The pod errored with the message: "Back-off pulling image "unleashorg/unleash-server:latest"."`

Will keep retrying - seems temporary.

Hyperkid123:
There are issues with Quay. Thats why we see the CI failures

Hyperkid123 (approved):

PR #776, opened 2025-01-31
fix: remove group under IAM and put service accounts where it should be

aferd:
[RHCLOUD-35954](https://issues.redhat.com/browse/RHCLOUD-35954)

service accounts should be under IAM section.


florkbr (approved):
Code looks good to me (compared to stage). Let's verify the issue is fixed once deployed to stage then we can deploy to prod.

PR #785, opened 2025-02-07
feat: Add Itless default layout for widgets

SteveHNH:
Add an Itless default widget layout. There are several widgets that are unavailable in the itless environments. This is meant to create a nicer default widget layout when those widgets are hidden or unavailable.

### DIFF
--- a/static/stable/itless/modules/fed-modules.json
+++ b/static/stable/itless/modules/fed-modules.json
@@ -9,7 +9,10 @@
                 "routes": [
                     {
                         "exact": true,
-                        "pathname": "/"
+                        "pathname": "/",
+                        "props": {
+                          "layoutType": "landingPageItless"
+                        }
                     }
                 ]
             }
--- a/rest/models/DashboardTemplate.go
+++ b/rest/models/DashboardTemplate.go
@@ -16,7 +16,8 @@ import (
 type AvailableTemplates string
 
 const (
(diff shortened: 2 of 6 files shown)


static/stable/itless/modules/fed-modules.json
```diff
@@ -9,7 +9,10 @@
                 "routes": [
                     {
                         "exact": true,
-                        "pathname": "/"
+                        "pathname": "/",
+                        "props": {
+                          "layoutType": "landingPageItless"
```
Hyperkid123, line 14: This is the right approach, we just have to make sure the landing page accepts this prop.

We should also add the same prop to the stage stable and prod stable modules configuration and give it the landing page layout as input before we start using it in the UI module itself.

SteveHNH: @Hyperkid123 landing page had the change merged into master today so it should be good to go. 
Should I just add this prop to the other environments in this PR or would you prefer a separate one?

Hyperkid123: It can be in this one. It will not break anything as it is not really used.

SteveHNH: @Hyperkid123 all set! Let me know if I need to change anything else.

Hyperkid123 (approved):

PR #789, opened 2025-02-11
Add global learning resources config

JetyAdam:


static/stable/itless/modules/fed-modules.json
ryelo, line 1: Same as above ^

static/beta/prod/modules/fed-modules.json
ryelo, line 1: I don't know if we want to add it to the itless envs yet

ryelo (dismissed):

static/stable/itless/modules/fed-modules.json
JetyAdam: Removed it from itless envs

static/stable/prod/modules/fed-modules.json
```diff
@@ -15,6 +15,22 @@
             }
         ]
     },
+    "globalLearningResourcesPage": {
+        "manifestLocation": "/apps/global-learning-resources/fed-mods.json",
```
florkbr, line 19: @JetyAdam I think we need to update the app-interface config to include this new app/route. On stage this doesn't return the fed-mods JSON: https://console.stage.redhat.com/apps/global-learning-resources/fed-mods.json

But for example widget-layout does: But for example this one does: https://console.stage.redhat.com/apps/widget-layout/fed-mods.json

florkbr: We also will need to update Akamai for this new app/route

florkbr: Actually wait, can't we just use `/apps/learning-resources/fed-mods.json` instead?

JetyAdam: We can use that. Updated it

static/beta/prod/modules/fed-modules.json
```diff
@@ -15,6 +15,22 @@
             }
         ]
     },
+    "globalLearningResourcesPage": {
+        "manifestLocation": "/apps/learning-resources/fed-mods.json",
+        "defaultDocumentTitle": "All Learning Resources",
+        "modules": [
+            {
+                "id": "globalLearningResourcesPage",
+                "module": "./RootApp",
```
florkbr, line 24: @JetyAdam I believe we want this to be `GlobalLearningResourcesPage` per https://github.com/RedHatInsights/learning-resources/blob/master/fec.config.js#L25

florkbr:
@JetyAdam I pushed a few commits and got things working locally to both your PRs. With the approach I took I had to remove your `"defaultDocumentTitle": "All Learning Resources",` change - so we should set the document title instead in the GlobalLearningResources module itself versus through the chrome configuration.

florkbr (approved):

PR #792, opened 2025-02-12
Missing appId and description for OpenShift Virtualization

davidaznaur:
Added missing appId and description for side nav OpenShift Virtualization Link


ryelo (approved):

static/beta/prod/navigation/openshift-navigation.json
florkbr, line 1: @davidaznaur just an FYI - these `/beta` files are no longer used and will be removed in the future.

PR #793, opened 2025-02-12
Place search result for global learning resources behind feature flag

florkbr:
Follow up to https://github.com/RedHatInsights/chrome-service-backend/pull/789

We need to place the search result for global learning resources behind the dark launch flag as well.


florkbr:
CC @JetyAdam

florkbr:
I'm going to make an exception and merge this PR and will validate it works on stage by disabling the feature flag for a moment. We need this to deploy global learning resources to prod.

PR #800, opened 2025-02-17
Fix generating duplicate nested nav items.

Hyperkid123:
There was a duplicate itteration in nested nav structures.


justinorringer (approved):
LGTM! Unrelated but the logic between those two if statements is exactly the same :sweat_smile: , removing duplicate code is always cool

PR #808, opened 2025-02-24
Add assisted-migration app

machacekondra:


Hyperkid123:
@machacekondra I've just checked and https://console.stage.redhat.com/apps/assisted-migration-app/fed-mods.json does not return any data. The configuration itself format is valid.

We can proceed with this PR and merge it, but the UI module will not work until the assets are exposed via our CDN.

ammont82:
> @machacekondra I've just checked and https://console.stage.redhat.com/apps/assisted-migration-app/fed-mods.json does not return any data. The configuration itself format is valid.
> 
> We can proceed with this PR and merge it, but the UI module will not work until the assets are exposed via our CDN.

This is the correct link: https://console.dev.redhat.com/apps/migration-assessment/fed-mods.json

ryelo (approved):

PR #809, opened 2025-02-26
RHCLOUD-36226 | feature: user's most recently used workspaces

MikelAlejoBR:
The feature allows saving and retrieving the most recently used workspaces by the console user.

## Jira ticket
[[RHCLOUD-36226]](https://issues.redhat.com/browse/RHCLOUD-36226)


go.mod
```diff
@@ -1,12 +1,12 @@
 module github.com/RedHatInsights/chrome-service-backend
 
-go 1.20
+go 1.21
```
Hyperkid123, line 3: Just make sure we have a build image for this version. Not sure if 21 is available yet.

MikelAlejoBR: I am testing things out, because I used the `slices` package for some things, but the tests are failing and I assume it's because of this. I'll try to revert every dependency upgrade back to what we had before and see if that solves the issues, and then figure out the `slices` and the `21` version thing.

MikelAlejoBR:
Finally! Rebasing the main branch worked.

go.mod
MikelAlejoBR: The image of the Dockerfile gets built no problem, so we're safe here! Sources uses `1.23` with `ubi9` as well, so I'd say we're safe here.

cmd/migrate/migrate.go
```diff
@@ -139,6 +139,14 @@ func main() {
 		panic(activeWorkspaceRes.Error)
 	}
 
+	fmt.Println("Set a default value to the recently used workspaces' column")
```
Hyperkid123, line 142: Debug line

rest/routes/recentlyUsedWorkspaces.go
```diff
@@ -0,0 +1,205 @@
...
+	// Try saving the most recently used workspaces in the database. Also, take into account the maximum number of
+	// recently used workspaces we want to save in the database.
+	var workspacesToSave []models.Workspace
+	if len(requestWorkspaces) > configuration.MaximumNumberRecentlyUsedWorkspaces {
+		workspacesToSave = requestWorkspaces[:configuration.MaximumNumberRecentlyUsedWorkspaces]
+	} else {
+		workspacesToSave = requestWorkspaces
+	}
+
+	if err := service.SaveRecentlyUsedWorkspaces(&user, workspacesToSave); err != nil {
```
Hyperkid123, line 181: We don't want doplicates in the array. I don't think you are checking for that. If we already have WS with matching id, it should just be moved to the top of the array.

cmd/migrate/migrate.go
MikelAlejoBR: Removed!

rest/routes/recentlyUsedWorkspaces.go
MikelAlejoBR: Added a deduplication mechanism. The duplicated workspaces are moved to the top of the list and deduplicated.

Hyperkid123 (approved):

PR #833, opened 2025-04-07
Fix make infra command in Makefile

JetyAdam:
This PR fixes issues with make infra and make clean-all commands in Makefile.


Makefile
```diff
@@ -62,11 +62,11 @@ unleash:
 	podman-compose -f local/unleash-compose.yaml up
 
 infra:
-	podman-compose -f local/full-stack-compose.yaml down
-	podman-compose -f local/full-stack-compose.yaml up
+	podman-compose -f $(PWD)/local/full-stack-compose.yaml down
+	podman-compose -f $(PWD)/local/full-stack-compose.yaml up
```
florkbr, line 66: I'd like to understand what changes to required this absolute path now (possibly a change in podman compose?)

If it is required, we will need to update the other commands above.

florkbr: @JetyAdam I was able to confirm - this is a bug in version 1.3.0 of `podman-compose`. See: https://github.com/containers/podman-compose/issues/1167

For now we can use PWD. @JetyAdam can you update the other podman-compose commands in this file?

florkbr (approved):

PR #834, opened 2025-04-07
Add view all link to bookmarked learning resources widget

JetyAdam:
https://issues.redhat.com/browse/RHCLOUD-38241

### DIFF
--- a/rest/models/DashboardTemplate.go
+++ b/rest/models/DashboardTemplate.go
@@ -16,7 +16,7 @@ import (
 type AvailableTemplates string
 
 const (
-	LandingPage AvailableTemplates       = "landingPage"
+	LandingPage       AvailableTemplates = "landingPage"
 	LandingPageItless AvailableTemplates = "landingPageItless"
 )
 
@@ -61,7 +61,7 @@ const (
 	Acs                 AvailableWidgets = "acs"
 	Subscriptions       AvailableWidgets = "subscriptions"
 	SupportCases        AvailableWidgets = "supportCases"
-	Integrations    	AvailableWidgets = "integrations"
+	Integrations        AvailableWidgets = "integrations"
 	ImageBuilder        AvailableWidgets = "imageBuilder"
 )
 
@@ -315,7 +315,7 @@ const (
 	ACSIcon              WidgetIcons = "ACSIcon"
 	OpenShiftAiIcon      WidgetIcons = "OpenShiftAiIcon"
 	HeadsetIcon          WidgetIcons = "HeadsetIcon"
-	IntegrationsIcon	 WidgetIcons = "IntegrationsIcon"
+	IntegrationsIcon     WidgetIcons = "IntegrationsIcon"
 )
 
 func (wi WidgetIcons) IsValid() error {
@@ -328,8 +328,9 @@ func (wi WidgetIcons) IsValid() error {
 }
 
 type WidgetHeaderLink struct {
-	Title string `json:"title,omitempty"`
-	Href  string `json:"href,omitempty"`
+	Title       string `json:"title,omitempty"`
+	Href        string `json:"href,omitempty"`
+	FeatureFlag string `json:"featureFlag,omitempty"`
 }
 
 type WidgetPermissionMethods string
@@ -355,12 +356,12 @@ type WidgetConfiguration struct {
 }
 
 type ModuleFederationMetadata struct {
-	Scope      string               `json:"scope"`
-	Module     string               `json:"module"`
-	ImportName string               `json:"importName,omitempty"`
-	FeatureFlag string              `json:"featureFlag,omitempty"`
-	Defaults   BaseWidgetDimensions `json:"defaults"`
-	Config     WidgetConfiguration  `json:"config"`
+	Scope       string               `json:"scope"`
+	Module      string               `json:"module"`
+	ImportName  string               `json:"importName,omitempty"`
+	FeatureFlag string               `json:"featureFlag,omitempty"`
+	Defaults    BaseWidgetDimensions `json:"defaults"`
+	Config      WidgetConfiguration  `json:"config"`
 }
 
 type WidgetModuleFederationMapping map[AvailableWidgets]ModuleFederationMetadata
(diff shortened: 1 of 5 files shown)


rest/service/dashboardTemplate.go
```diff
@@ -162,6 +162,10 @@ var (
 			Defaults: models.BaseWidgetDimensions.InitDimensions(models.BaseWidgetDimensions{}, 2, 4, 10, 1),
 			FeatureFlag: "widget.learningResources.hidden",
 			Config: models.WidgetConfiguration{
+				HeaderLink: models.WidgetHeaderLink{
+					Title: "View all",
+					Href:  "/learning-resources?tab=all",
```
florkbr, line 167: @JetyAdam FWIW this change will not be behind the feature flag we've created. Unfortunately the `WidgetHeaderLink` model doesn't contain feature flag support at the moment: https://github.com/RedHatInsights/chrome-service-backend/blob/main/rest/models/DashboardTemplate.go#L330-L333

Since we are enabling the flag ASAP we can coordinate the rollout of this PR and enabling the flag for all - however in practice in the future I'd like to make sure the feature is completely hidden behind the flag in case of any issues.

florkbr (changes requested):
@JetyAdam I confirmed we do not have global learning resources deployed in itless, **but we do have the bookmarked learning resources widget** - so we will need to hide this behind the feature flag - otherwise there will be a broken link in the widget in itless.

See an example here: https://github.com/RedHatInsights/chrome-service-backend/blob/main/main.go#L76

rest/service/dashboardTemplate.go
JetyAdam: Updated. I hid the link behind a feature flag.

rest/models/DashboardTemplate.go
```diff
@@ -16,7 +16,7 @@ import (
 type AvailableTemplates string
 
 const (
-	LandingPage AvailableTemplates       = "landingPage"
+	LandingPage       AvailableTemplates = "landingPage"
```
florkbr, line 19: @JetyAdam I checked locally and my IDE (vscode) also modifies these whitespaces - but only when I modify the file. Since we don't have any logic changes in this file, can we just reset this file with upstream to avoid merge conflicts?

rest/routes/dashboardTemplate.go
```diff
@@ -272,12 +272,36 @@ func DecodeDashboardTemplate(w http.ResponseWriter, r *http.Request) {
 	handleDashboardResponse[models.DashboardTemplate](resp, err, w)
 }
 
+func deepCopyJSON(metadata models.ModuleFederationMetadata) (models.ModuleFederationMetadata, error) {
+	data, err := json.Marshal(metadata)
+	if err != nil {
+		return models.ModuleFederationMetadata{}, err
```
florkbr, line 278: @JetyAdam I was thinking we have 2 options here and the error handle below (both are not really ideal):
1. Return an empty widget metadata like we do here
2. Return the original widget metadata (unmodified)

If we return empty metadata it will be explicitly obvious something went wrong with this widget when performing our modification based on feature flags. The entire widget will be omitted from the mapping. If we return the original - we may be showing more than we intend as the feature flag check failed.

In practice that should hopefully be rare - but it's something to consider. 

If we do this check in the UI - we can be a bit more flexible on error handling. @Hyperkid123 thoughts?

rest/routes/dashboardTemplate.go
```diff
@@ -272,12 +272,36 @@ func DecodeDashboardTemplate(w http.ResponseWriter, r *http.Request) {
...
+	}
+	return newMetadata, nil
+}
+
+func FilterWidgetMappingHeaderLink(widgetMapping models.WidgetModuleFederationMapping) models.WidgetModuleFederationMapping {
+	for key, value := range widgetMapping {
+		if value.Config.HeaderLink.FeatureFlag != "" && !featureflags.IsEnabled(value.Config.HeaderLink.FeatureFlag) {
+			deepCopy, _ := deepCopyJSON(widgetMapping[key])
```
florkbr, line 291: If we do keep this approach - we should handle the error (`_`) case

rest/routes/dashboardTemplate.go
JetyAdam: I lean toward the first option, since we want to avoid any unintended side effects.

rest/routes/dashboardTemplate.go
JetyAdam: Updated the error case. In the error case, I remove the header link from the original map instead of the deep copy.

rest/routes/dashboardTemplate.go
florkbr: @JetyAdam do you mind adding tests then? I want to make sure we cover all the expected cases and error behavior is predictable. We can use `recentlyUsedWorkspaces_test.go` test already in this directory as an example.

JetyAdam: Added the test

Hyperkid123: As all of this will be eventually handed by FEO I am an fine with whatever you decide right now.

rest/routes/deepCopyJSON_test.go
```diff
@@ -0,0 +1,53 @@
+package routes
+
+import (
+	"testing"
+
+	"github.com/RedHatInsights/chrome-service-backend/rest/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDeepCopyJSON(t *testing.T) {
```
florkbr, line 10: @JetyAdam we are missing tests for the `FilterWidgetMappingHeaderLink` function - which I think is the riskier function as it is performing the modifications on the mapped widget object.

Since this PR has been out there for some time I'll confirm everything locally - but I'd like to follow up with another test

rest/routes/deepCopyJSON_test.go
florkbr, line 1: @JetyAdam the name of this file should match the name of the file we are testing - so `dashboardTemplate_test.go`

florkbr:
Confirmed locally by running chrome service with these changes and chrome UI and manipulating the feature flags. Works as intended :shipit:

PR #835, opened 2025-04-08
Update widget layout logic

SteveHNH:
Create a new optional boolean within the ModuleFederationMetadata that specifies the widget should be used in an itless environment. This makes every widget off by default rather than on, making management of the widgets easier.

It also moves the check to the FRONTEND_ENVIRONMENT var rather than a feature flag

### DIFF
--- a/rest/routes/dashboardTemplate.go
+++ b/rest/routes/dashboardTemplate.go
@@ -285,7 +285,6 @@ func FilterWidgetMapping(widgetMapping models.WidgetModuleFederationMapping) mod
 func GetWidgetMappings(w http.ResponseWriter, r *http.Request) {
 	var err error
 	var resp util.EntityResponse[models.WidgetModuleFederationMapping]
-	logrus.Errorln(err)
 
 	if featureflags.IsEnabled("chrome-service.filterWidgets.enable") {
 		resp = util.EntityResponse[models.WidgetModuleFederationMapping]{
--- a/rest/routes/dashboardTemplate.go
+++ b/rest/routes/dashboardTemplate.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
-	"os"
 	"strconv"
 
 	"github.com/RedHatInsights/chrome-service-backend/rest/featureflags"
--- a/rest/routes/dashboardTemplate.go
+++ b/rest/routes/dashboardTemplate.go
@@ -322,6 +322,7 @@ func GetWidgetMappings(w http.ResponseWriter, r *http.Request) {
 
 		resp = util.EntityResponse[models.WidgetModuleFederationMapping]{
 			Data: filteredWidgetMapping,
+		}
 	} else {
 		resp = util.EntityResponse[models.WidgetModuleFederationMapping]{
 			Data: service.WidgetMapping,
--- a/rest/routes/dashboardTemplate.go
+++ b/rest/routes/dashboardTemplate.go
@@ -272,6 +272,36 @@ func DecodeDashboardTemplate(w http.ResponseWriter, r *http.Request) {
 	handleDashboardResponse[models.DashboardTemplate](resp, err, w)
 }
 
+func deepCopyJSON(metadata models.ModuleFederationMetadata) (models.ModuleFederationMetadata, error) {
+	data, err := json.Marshal(metadata)
+	if err != nil {
+		return models.ModuleFederationMetadata{}, err
+	}
+	var newMetadata models.ModuleFederationMetadata
+	err = json.Unmarshal(data, &newMetadata)
+	if err != nil {
+		return models.ModuleFederationMetadata{}, err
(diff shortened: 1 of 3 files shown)


rest/routes/dashboardTemplate.go
```diff
@@ -272,12 +272,10 @@ func DecodeDashboardTemplate(w http.ResponseWriter, r *http.Request) {
 	handleDashboardResponse[models.DashboardTemplate](resp, err, w)
 }
 
-
-// FilterWidgetMapping removes hidden widgets from the mapping by using feature flags stored with the widget definition.
 func FilterWidgetMapping(widgetMapping models.WidgetModuleFederationMapping) models.WidgetModuleFederationMapping {
 	for key, value := range widgetMapping {
-		if featureflags.IsEnabled(value.FeatureFlag) {
-			delete(widgetMapping, key)	
+		if !value.Itless {
```
florkbr, line 277: @SteveHNH putting some thought into the problem of: **new widget unexpectedly shows up in itless.**

Could we in theory still use feature flags for this, but make the itless default behavior a bit different?

I was thinking:
* **For both commercial and itless** we support conditional rendering of the widgets based on feature flags (optional opt-in)
* **For both commercial and itless** if the feature flag is provided and at runtime is set to "false", we remove the widget. If provided and set to "true" at runtime, we show the widget. 
* **In commercial** - If no feature flag value is provided (IE empty string `""`) on the metadata - **we do nothing**. This means newly added widgets will show up by default.
* **In itless** - If no feature flag value is provided (IE empty string `""`) on the metadata - **we remove the widget**. This means newly added widgets will be removed by default.

So really the only way to get a widget to show in itless would be to specify a feature flag, then enable it at runtime.

**Drawback**: whatever feature flag we choose for the widgets that do show up in itless would also need to be set to true in commercial instances - though the flag itself is not environment specific.

I realize that seems odd and maybe more over-engineered. My idea is to minimize the amount of itless specific logic required in source for longer term maintainability. Feature flags are used in both envs - just differing behavior. We'd just need to be able to tell we are running in an itless env to do that (which we can also use an feature flag for).

SteveHNH: Yeah, I like it. let me work on it a bit and I'll get this updated.

SteveHNH:
@florkbr ALright, made some changes. If FeatureFlag is empty  or missing, we go ahead and show the widget. For added protection, we only filter the widgets if we have a feature flag set for that, otherwise we just show them all. So this won't impact commercial at all unless they decide to do it. We've effectively just switched the logic from selecting what to hide in itless, to selecting what we want to be shown. Given that what we show is a small subset, this makes less work

SteveHNH:
@Hyperkid123 thanks for the retest. I'll get those errors cleaned up

SteveHNH:
@florkbr Resurrecting this thing. I think it should be all set and I cleared up _most_ of the errors. There's one konflux error but I don't think it has anything to do with this change.

rest/service/dashboardTemplate.go
```diff
@@ -140,8 +133,8 @@ var (
 		models.NotificationsEvents: models.ModuleFederationMetadata{
 			Scope:       "notifications",
 			Module:      "./DashboardWidget",
+			FeatureFlag: "widget.notifications.enabled",
```
florkbr, line 136: @SteveHNH noticed this flag is different than the rest - note `enabled` vs `enable` postfix. Is that intentional?

rest/routes/dashboardTemplate.go
```diff
@@ -316,13 +316,14 @@ func FilterWidgetMapping(widgetMapping models.WidgetModuleFederationMapping) mod
 func GetWidgetMappings(w http.ResponseWriter, r *http.Request) {
 	var err error
 	var resp util.EntityResponse[models.WidgetModuleFederationMapping]
+	logrus.Errorln(err)
```
florkbr, line 319: @SteveHNH we should make sure err is set/defined before we try to log anything

rest/service/dashboardTemplate.go
SteveHNH: This was an oversight. Good eye.

florkbr:
@SteveHNH the failing EC check was resolved in main if you pull!

florkbr (approved):
We currently do not enable `chrome-service.filterWidgets.enable` in commercial so this is safe to merge.

If we ever _do_ want to enable it in commercial (and I do see value there for dark launching new widgets) - we'd need to:
* enable that flag
* implement the `if no feature flag value` provided logic I talk about here: https://github.com/RedHatInsights/chrome-service-backend/pull/835#discussion_r2049355768, otherwise all the existing commercial widgets with no flags will disappear. 
* update widget developer documentation

I'll create an issue in our backlog to track those enhancements to this logic.

CC @Hyperkid123 we should pull this into https://github.com/RedHatInsights/widget-layout-backend

PR #854, opened 2025-05-07
Add roadmap to services dropdown

hosekadam:
https://issues.redhat.com/browse/RHCLOUD-38610

## Summary by Sourcery

New Features:
- Added roadmap as a new option in the services dropdown

### DIFF
--- a/static/stable/stage/services/services.json
+++ b/static/stable/stage/services/services.json
@@ -372,8 +372,9 @@
         "title": "Planning",
         "links": [
           {
-            "href": "/insights/roadmap/lifecycle",
-            "title": "Lifecycle",
+            "href": "/insights/planning/lifecycle",
+            "title": "Life Cycle",
+            "id": "roadmap.lifecycle",
             "description": "View tailored life cycle data for Red Hat Enterprise Linux and RHEL application streams.",
             "alt_title": [
               "planning",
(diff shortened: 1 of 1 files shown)


static/stable/stage/services/services.json
```diff
@@ -366,7 +366,41 @@
             "href": "/subscriptions/usage/openshift"
           }
         ]
-      }
+      },
+      {
+        "isGroup": true,
```
tahmidefaz, line 371: I think this  is missing an `id` key?

static/stable/stage/services/services.json
```diff
@@ -503,6 +537,28 @@
           "application-services.trustedContentOverview",
           "application-services.trustedArtifactSignerOverview"
         ]
+      },
+      {
+        "isGroup": true,
```
tahmidefaz, line 542: same here as well?

tahmidefaz (changes requested):
Please check the comments above. Verified, these needs the `id` key.

hosekadam:
ID keys added, applied some suggestions from John Carr. 

But probably this PR can be closed and used as data source for the FEO (as the application is already moved there).

tahmidefaz:
@hosekadam We actually need the `id` key for each group as well as for each section/links. For example: [Here](https://github.com/RedHatInsights/chrome-service-backend/blob/main/static/stable/stage/services/services.json#L510-L519)

tahmidefaz:
> ID keys added, applied some suggestions from John Carr.
> 
> But probably this PR can be closed and used as data source for the FEO (as the application is already moved there).

Sounds good! Let us know if this one needs a merge.

PR #853, opened 2025-05-07
Add Roadmap search results

hosekadam:
https://issues.redhat.com/browse/RHCLOUD-38609

## Summary by Sourcery

New Features:
- Introduced Roadmap search results in the static services entries

PR #856, opened 2025-05-07
Fix roadmap title

hosekadam:
Follow up of #853 based on review from John Carr.

## Summary by Sourcery

Chores:
- Modify static services entries based on previous review

### DIFF
--- a/cmd/search/static-services-entries.json
+++ b/cmd/search/static-services-entries.json
@@ -1004,7 +1004,7 @@
       {
         "custom": true,
         "id": "roadmap.lifecycle",
-        "href": "/insights/roadmap/lifecycle",
+        "href": "/insights/planning/lifecycle",
         "title": "Life Cycle",
         "description": "View tailored life cycle data for Red Hat Enterprise Linux and RHEL application streams.",
         "alt_title": [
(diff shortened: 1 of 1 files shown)


cmd/search/static-services-entries.json
```diff
@@ -1004,8 +1004,8 @@
       {
         "custom": true,
         "id": "roadmap.lifecycle",
-        "href": "/insights/roadmap/lifecycle",
-        "title": "Lifecycle",
+        "href": "/insights/planning/lifecycle",
```
hosekadam, line 1007: There is probably another thing happening, as we are moving from the `fed`config and doing feo migration, do we want to have the #853 #854 and #856 for chrome service, or it should be plased elsewhere?

hosekadam: @Hyperkid123  No need for merging this PR, the migration already happened, and the result is here https://github.com/RedHatInsights/digital-roadmap-frontend/blob/main/deploy/frontend.yaml (if you want to take a look). I will create PR for reverting #853 so it's not duplicit.

Hyperkid123: Kudos on using the CR!

PR #866, opened 2025-05-22
fix(RHINENG:16922): Update Remediations title for breadcrumbs

adonispuente:
This simply updates the title so that breadcrumbs will now match the nav bar


tahmidefaz (approved):
lgtm!

PR #880, opened 2025-06-04
fix: use partial index on the user_identities table account id field

florkbr:
Follow up from our investigations into slow DB queries.

This will add a partial index on the account_id in the user_identities table (where deleted_at is null).

~We use gorm autoMigrate functionality - so this annotation should be picked up and applied to the database when the migrate script is run (on pod init): https://github.com/RedHatInsights/chrome-service-backend/blob/main/cmd/migrate/migrate.go#L102~

~Setting as draft so I can test a few things locally first.~

Edit: if we want to use `concurrently` this cannot be done within an existing transaction/automigrate - so I've shifted to creating the index through executing SQL in the migrate script (after the automigrate completes). This can also be done in app-interface, but I wanted to stick to using our migrate script.

## Summary by Sourcery

Enhancements:
- Add GORM partial index on the user_identities table’s account_id field to only index non-deleted records

### DIFF
--- a/cmd/migrate/migrate.go
+++ b/cmd/migrate/migrate.go
@@ -170,5 +170,14 @@ func main() {
 		logrus.Infof("Migrated %d user identity visited rows", activeWorkspaceRes.RowsAffected)
 	}
 
+	// must be outside of a transaction (to support concurrently creating the index)
+	fmt.Println("Concurrently create partial index for user_identities table")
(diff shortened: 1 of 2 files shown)


florkbr:
@Hyperkid123 can we deploy this to stage in your AM and monitor the results?

Hyperkid123:
@florkbr lest do it.

Hyperkid123 (approved):

PR #897, opened 2025-07-08
Migrate repo to use shared Konflux pipelines

JetyAdam:
https://issues.redhat.com/browse/RHCLOUD-39550

## Summary by Sourcery

Enhancements:
- Replace embedded pipelineSpec blocks in pull-request and push YAMLs with a pipelineRef pointing to the centralized konflux-pipelines repository.


.tekton/chrome-service-pull-request.yaml
```diff
@@ -27,660 +27,15 @@ spec:
...
-          name: validate-schemas-files
-          script: "#!/bin/bash\nset -ex\nmake validate-schema\nmake publish-search-index-dry-run\n make parse-services\nmake generate-search-index  \n"
-          securityContext:
-            runAsUser: 0
-          workingDir: /var/workdir/source
-        volumes:
-        - emptyDir: {}
-          name: workdir
-    - description: Validates chrome configuration files
-      name: run-unit-tests
```
justinorringer, line 294: This isn't going to be included in the shared pipelines, we should add it to github actions.

justinorringer (changes requested):
Adding the pipeline looks good, we should recreate the unit tests in a github actions

.tekton/chrome-service-pull-request.yaml
Hyperkid123: There should be a way to extend the pipeline with tests (or for the frontend one, there is). But yes, we can't lose these.

Hyperkid123 (changes requested):
The tests and other checks have to run in Konflux. There is a way to extend the remote pipeline. If this one in particular does not support that, we need to write a new one for our services where we know we have this requirement.

Migrating to GH actions is not an option when we suitable environment in Konflux.

.tekton/chrome-service-pull-request.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: '3'
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main"
+    pipelinesascode.tekton.dev/pipeline: https://github.com/RedHatInsights/konflux-pipelines/raw/main/pipelines/platform-ui/docker-build-run-unit-tests.yaml
```
florkbr, line 11: @JetyAdam this is what is causing the failures - Go is not installed as part of the UI builds. Can we switch to the non-ui docker build?

florkbr: If the non-ui docker build does not run tests, we may need to update the shared pipelines to have an option to do so

.tekton/chrome-service-pull-request.yaml
```diff
@@ -27,665 +28,70 @@ spec:
...
+  - name: PGSQL_DATABASE
+    value: db
+  - name: UNLEASH_API_TOKEN
+    value: default:development.unleash-insecure-api-token
+  - name: UNLEASH_ADMIN_TOKEN
+    value: '*:*.unleash-insecure-api-token'
+  - name: unit-tests-script
+    value: |
+      #!/bin/bash
+      set -ex
```
gwenneg, line 50:
```suggestion
      set -euxo pipefail
      id -u
      whoami || true
      test -w /var/lib/rpm && echo "rpmdb writable" || echo "rpmdb NOT writable"
      test -w / && echo "/ writable" || echo "/ NOT writable"
```

gwenneg, line 50:
```suggestion
      set -euxo pipefail
      id -u
      whoami || true
      test -w /var/lib/rpm && echo "rpmdb writable" || echo "rpmdb NOT writable"
      test -w / && echo "/ writable" || echo "/ NOT writable"
```

Hyperkid123 (approved):

.tekton/chrome-service-push.yaml
```diff
@@ -24,665 +25,76 @@ spec:
...
+      export UNLEASH_ADMIN_TOKEN="$(params.UNLEASH_ADMIN_TOKEN)"
+      
+      make migrate
+      make test
+  pipelineRef:
+    name: docker-build
+  taskRunTemplate:
+    serviceAccountName: build-pipeline-chrome-service
+    podTemplate:
+      env:
```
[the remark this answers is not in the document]
JetyAdam: Nice catch! I removed it

PR #934, opened 2025-08-20
migrating to project hosted navigation system

zwitter1:
Trying to migrate iam console to local navigation files

## Summary by Sourcery

Migrate the IAM console to use project-hosted navigation and service configuration files

Enhancements:
- Relocate IAM navigation JSON to local project files for prod and stage environments
- Update service definitions in local JSON for both prod and stage

### DIFF
--- a/static/stable/stage/navigation/iam-navigation.json
+++ b/static/stable/stage/navigation/iam-navigation.json
@@ -202,8 +202,10 @@
       ]
     },
     {
+      "id": "authentication-policy",
       "title": "Authentication Policy",
       "expandable": true,
+      "feoReplcement": "authentication-policy",
       "routes": [
         {
           "id": "authFactors",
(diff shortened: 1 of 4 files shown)


static/stable/stage/navigation/iam-navigation.json
```diff
@@ -202,8 +202,10 @@
       ]
     },
     {
+      "id": "authentication-policy",
       "title": "Authentication Policy",
       "expandable": true,
+      "feoReplcement": "authentication-policy",
```
florkbr, line 208: @zwitter1 FWIW you'll probably need to add this to the prod config in the future as well. You can add it before hand and nothing should happen until you deploy IAM to prod with the `feoConfigEnabled`/updated frontend CRD. Just an FYI

florkbr (approved):

PR #957, opened 2025-09-04
fix(parseServices): Fix local development errors map of undefined on links in services

florkbr:
This is causing one of the errors in the React error boundary locally during development. We've removed the offending service link entry but also updated the script to handle this in the future (will return empty array now in lieu of null)

## Summary by Sourcery

Fix local development errors caused by missing service links by removing the invalid link entry and returning an empty links array instead of null

Bug Fixes:
- Initialize service item links as an empty slice to avoid nil dereference errors
- Remove the offending service link entry from the static services JSON


justinorringer (approved):
LGTM

PR #1090, opened 2026-05-14
Add GET /api/chrome-service/v1/api-docs endpoint

radekkaluzik:
## Summary

Add a new authenticated endpoint `GET /api/chrome-service/v1/api-docs` that serves aggregated OpenAPI specifications to the frontend Help Panel component.

## Changes

- **rest/routes/apiDocs.go** — New route handler that reads `static/specs-generated.json` from disk and returns raw JSON content. Returns `{}` if the file has not been generated yet.
- **rest/routes/apiDocs_test.go** — Unit tests covering success path, file-not-found fallback, Content-Type header, and verification that the response is not wrapped in an EntityResponse envelope.
- **main.go** — Register `/api-docs` route in the authenticated subrouter.

## Context

The CronJob implemented in RHCLOUD-37864 already generates `specs-generated.json` daily on stage. This PR adds the HTTP endpoint that exposes that data to the frontend. The response is served as raw JSON with no deserialization or transformation — the file content is streamed directly to the client.

## Related

- Jira: https://redhat.atlassian.net/browse/RHCLOUD-47439
- CronJob PR: https://github.com/RedHatInsights/chrome-service-backend/pull/1054

---

### Checklist
- [x] Tests: new/updated tests cover the change
- [x] API: spec updated if endpoints changed (or N/A)
- [x] Migrations: backwards compatible if schema changed (or N/A)
- [x] Dependencies: no known impact to dependent services
- [x] Security: reviewed against [secure coding checklist](https://github.com/RedHatInsights/secure-coding-checklist) (or N/A)

## Summary by Sourcery

Expose an authenticated API endpoint that serves pre-generated OpenAPI documentation as raw JSON to clients.

New Features:
- Add a new authenticated /api-docs route that streams the contents of static/specs-generated.json without additional wrapping.

Tests:
- Introduce unit tests verifying successful spec retrieval, file-not-found fallback behavior, JSON content type, and absence of response envelope wrapping.

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,6 +69,7 @@ func main() {
 		subrouter.Route("/user", routes.MakeUserIdentityRoutes)
 		subrouter.Route("/emit-message", routes.BroadcastMessage)
 		subrouter.Route("/dashboard-templates", routes.MakeDashboardTemplateRoutes)
+		subrouter.Route("/api-docs", routes.MakeApiDocsRoutes)
 	})
 
 	// We might want to set up some event listeners at some point, but the pod will
--- a/rest/routes/apiDocs.go
+++ b/rest/routes/apiDocs.go
@@ -1,0 +1,93 @@
+package routes
+
+import (
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"os"
+	"slices"
+
+	"github.com/RedHatInsights/chrome-service-backend/rest/util"
+	"github.com/go-chi/chi/v5"
+	"github.com/sirupsen/logrus"
+)
+
+const specsFilePath = "static/specs-generated.json"
+
+type apiDocEntry struct {
+	URL          string          `json:"url"`
+	BundleLabels []string        `json:"bundleLabels"`
+	Spec         json.RawMessage `json:"spec"`
+}
+
+func GetApiDocs(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Cache-Control", "max-age=3600")
+
+	bundleFilter := r.URL.Query().Get("bundle")
+
+	if bundleFilter == "" {
+		serveSpecsFile(w)
+		return
+	}
+
+	serveFilteredSpecs(w, bundleFilter)
+}
+
+func serveSpecsFile(w http.ResponseWriter) {
+	f, err := os.Open(specsFilePath)
+	if err != nil {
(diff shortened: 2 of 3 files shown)


karelhala (commented):
Looks good overall — clean handler, solid tests, and the streaming approach is the right call.

One thing worth considering before merging: the APIs Help Panel will support grouping by bundle (same as Learn does today). Adding a `?bundle=settings` query parameter now would let the frontend integrate with the final API shape from day one instead of fetching everything and filtering client-side. If no `bundle` param is provided, the endpoint should return all bundles (current behavior).

Other minor nits:

1. **`w.Write` error on fallback path** — the `w.Write([]byte("{}"))` on the not-found path silently discards the error, while the `io.Copy` path logs it. Minor inconsistency.

2. **No caching headers** — since the specs file is generated daily, a `Cache-Control: max-age=3600` or similar could reduce repeated fetches from the Help Panel. Fine as a follow-up.

radekkaluzik:
> Looks good overall — clean handler, solid tests, and the streaming approach is the right call.
> 
> One thing worth considering before merging: the APIs Help Panel will support grouping by bundle (same as Learn does today). Adding a `?bundle=settings` query parameter now would let the frontend integrate with the final API shape from day one instead of fetching everything and filtering client-side. If no `bundle` param is provided, the endpoint should return all bundles (current behavior).
> 
> Other minor nits:
> 
> 1. **`w.Write` error on fallback path** — the `w.Write([]byte("{}"))` on the not-found path silently discards the error, while the `io.Copy` path logs it. Minor inconsistency.
> 2. **No caching headers** — since the specs file is generated daily, a `Cache-Control: max-age=3600` or similar could reduce repeated fetches from the Help Panel. Fine as a follow-up.

@karelhala All three review items have been addressed:

1. **Bundle filtering** — Added `?bundle=` query parameter support. When provided (e.g. `?bundle=settings`), the endpoint returns only specs whose `bundleLabels` contain the requested bundle. When omitted, returns all specs (previous behavior preserved via raw file streaming).

2. **Fallback write error handling** — Extracted shared `handleFileError()` function that logs errors on both the not-found fallback path and the I/O error path consistently.

3. **Cache-Control header** — Added `Cache-Control: max-age=3600` to all response paths.

Three new tests added: bundle filtering (match + no match) and Cache-Control header verification. All 8 tests passing.

justinorringer (approved):
LGTM!

PR #1127, opened 2026-06-24
fix: remove dead V1 workspaces navigation

CodyWMitchell:
### Description

[RHCLOUD-47873
](https://redhat.atlassian.net/browse/RHCLOUD-47873)

- The `platform.rbac.workspaces-list` feature flag is no longer being used, so the V1 workspaces nav should be removed.
- Fixes issue with workspaces nav showing showing in V1 (where `platform.rbac.workspaces` flag set to false).

---

### How to test locally

---

### Anything reviewers should know?

---

### Checklist
- [ ] Tests: new/updated tests cover the change
- [ ] API: spec updated if endpoints changed (or N/A)
- [ ] Migrations: backwards compatible if schema changed (or N/A)
- [ ] Dependencies: no known impact to dependent services
- [ ] Security: reviewed against [secure coding checklist](https://github.com/RedHatInsights/secure-coding-checklist) (or N/A)

### AI disclosure

Assisted by: Claude Code

## Summary by Sourcery

Remove deprecated V1 workspaces navigation entries from IAM navigation configs.

Bug Fixes:
- Prevent legacy V1 workspaces navigation from appearing when the new workspaces feature flag is disabled.

Enhancements:
- Align IAM navigation configurations across envs to rely solely on the current workspaces feature flag.

### DIFF
--- a/static/beta/prod/navigation/iam-navigation.json
+++ b/static/beta/prod/navigation/iam-navigation.json
@@ -97,23 +97,6 @@
           "product": "Identity & Access Management"
         },
         {
-          "id": "workspaces",
-          "appId": "rbac",
-          "title": "Workspaces",
-          "icon": "PlaceholderIcon",
-          "href": "/iam/user-access/workspaces",
-          "permissions": [
-            {
-              "method": "featureFlag",
-              "args": ["platform.rbac.workspaces-list", true]
-            },
-            {
-              "method": "isOrgAdmin"
-            }
-          ],
-          "product": "Identity & Access Management"
-        },
-        {
           "id": "accessRequests",
           "appId": "rbac",
           "title": "Red Hat Access Requests",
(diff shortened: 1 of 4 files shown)


riccardo-forina (approved):

riccardo-forina:
@RedHatInsights/experience-services I'm trying to understand if the behavior we see in prod is correct. insights-rbac-ui uses the FEO config to handle the navigation, but it looks like it's being merged with this old config. Is that right? I vaguely remember setting up rbac-ui to not to do any merging, but maybe we just chatted about this

justinorringer:
Hey! if you all add your app to this list, it wont use any legacy nav from here. https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/chrome-service/deploy.yml?ref_type=heads#L99

that is already enabled in stage for iam bundle

